PR #35, opened 2015-02-04
JC-2016 New draft validation policy

mixas27:
Now drafts can be saved with only one field filled (not matter recipient, title or body):
      -Created separate DTO for drafts to perform specific validation rules

Added suggestion list to recipient field. Now if user types invalid recipient name recipient field will be cleared after leaving focus or before submitting form(if user press ctl + enter hotkey)

Created unit tests using Spring MockMvc

### DIFF
--- a/jcommune-view/jcommune-web-controller/src/main/java/org/jtalks/jcommune/web/validation/validators/ValidatorStub.java
+++ b/jcommune-view/jcommune-web-controller/src/main/java/org/jtalks/jcommune/web/validation/validators/ValidatorStub.java
@@ -1,0 +1,47 @@
+/**
+ * Copyright (C) 2011  JTalks.org Team
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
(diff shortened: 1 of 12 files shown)


jcommune-view/jcommune-web-controller/src/main/java/org/jtalks/jcommune/web/validation/validators/ValidatorStub.java
```diff
@@ -0,0 +1,47 @@
...
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ */
+package org.jtalks.jcommune.web.validation.validators;
+
+import org.springframework.validation.Errors;
+import org.springframework.validation.Validator;
+
+/**
+ * Simple implementation of {@link org.springframework.validation.Validator} interface.
```
ctapobep: Intention is not clear, could you please tell more about when and how it's to be used?

PR #36, opened 2015-02-05
JC-2103 HEAD not supported exception for root page (/)

standpoint:
Filter-mapping extends to all possible roots


mixas27:
Merged

PR #38, opened 2015-02-16
Jc 2073 create edit delete comments

NCNecros:


jcommune-model/src/main/java/org/jtalks/jcommune/model/entity/PostComment.java
```diff
@@ -152,4 +153,22 @@ public void putAttribute(String attributeName, String attributeValue) {
     public void putAllAttributes(Map<String, String> attributes) {
         this.attributes.putAll(attributes);
     }
+
+    /**
+     * Get a sign of delete
+     *
+     * @return true if the comment has been deleted
+     */
+    public Boolean getDeleted() {
```
mixas27: Primitive should be used here

jcommune-model/src/main/java/org/jtalks/jcommune/model/entity/PostComment.java
```diff
@@ -152,4 +153,22 @@ public void putAttribute(String attributeName, String attributeValue) {
...
+        return deleted;
+    }
+
+    /**
+     * Set a sign of delete to the comment
+     *
+     * @param deleted
+     */
+    public void setDeleted(Boolean deleted) {
```
mixas27: again

mixas27:
What about hibernate mapping? And don't forget unit tests

jcommune-model/src/main/java/org/jtalks/jcommune/model/entity/PostComment.java
ctapobep: @mixas27 what do you think about using `DATE_DELETED` of DateTime type instead of boolean?

mixas27: Will it be useful?

ctapobep: For auditing purposes.. We'll know when it happened. In the future we might add the author.. 

Or we could try using Envers right now: http://docs.jboss.org/envers/docs/ This helper-framework for Hibernate will manage deleting/editing logic on its own. But I have no experience with it, so can't say whether it has hidden issues.

mixas27: If it will be used by somebody, why not

NCNecros: What about use isDeleted instead getDeleted? It is very talking name.

jcommune-model/src/main/java/org/jtalks/jcommune/model/entity/PostComment.java
```diff
@@ -152,4 +153,22 @@ public void putAttribute(String attributeName, String attributeValue) {
     public void putAllAttributes(Map<String, String> attributes) {
         this.attributes.putAll(attributes);
     }
+
+    /**
+     * Get a sign of delete
+     *
+     * @return true if the comment has been deleted
+     */
+    public boolean getDeleted() {
```
mixas27: Yes, following conventions, it should be ```isDeleted```

jcommune-model/src/main/resources/org/jtalks/jcommune/migrations/V65__Add_column_DELETED_to_Post_Comment.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE POST_COMMENT ADD(DELETED TINYINT(1) NOT NULL DEFAULT 0);
```
mixas27: We discussed with @ctapobep and decided to use datetime of deletion instead of flag.

PR #42, opened 2015-02-23
JC-1573 Incorrect page makeup in messages when you insert link

wertklop:
Deleted 2 attributes of class "dialog-info". This not influence on working app.

Same as PR jtalks-org/jcommune/pull/41


wertklop:
Если я сейчас удалю ветку JC-1573, это ни на что не повлияет?

ctapobep:
Nope, it won't effect. The changes is already merged into upstream.

PR #50, opened 2015-03-01
JC-1669 Wrong page title when validation error happens during topic creation

wertklop:
Added validation of branch on NULL fo code review creation


ctapobep:
Could you also check the questions-n-answers plugin as well? After you build the jar, you can place it inside your home directory, by default plugins are read from there.

wertklop:
In questions-n-answers plugin all right and these changes don't affect in plugin behavior

PR #51, opened 2015-03-01

wertklop:
JC-2087 Lower-case text in message error

Added message error with uppercase beginning of the hints in new article page (EN, SPA). Both issues are equals


ctapobep:
@wertklop what do you think about getting rid of _May not be empty_ and instead always show _Size should be ..._? It seems to me that these empty-error messages are not that informative..

wertklop:
May be you are right, but error messages in issue has that text messages. If need other texts, i change it. Give me concrete text of messages, please

ctapobep:
Ok, I've added a JIRA comment to consolidate all the error messages on all the forms.

ctapobep:
@wertklop should I reassign this task to another person?

ctapobep:
Was finished in different PRs.

PR #53, opened 2015-03-02
JC-2144 Sample forum script shouldn't create admin user any more + bug fixes

begemosha:
1. admin/admin user creation removed, because it always created by new
versions of JC at first start and attempt to create it led to bugs,
particularly JC-2144.
2. Script ecpected, that all acl_ tables are empty. This assumption is
wrong nowadays and script created only part of intended branch rights
and etc. Fixed.
3. Duplication in order of Branches removed. There was too branches with
order = 9.
4. FORUM component creation replaced to update of description. Component
is always exists before script run and creation used to fail.

We shoudn't forget to update sample forum script when change migration scripts!


ctapobep:
Thanks! Hope the script will finally serve us good :)

begemosha:
I've checked it on major deploy scenarious before commiting.

begemosha:
But why it's not merged...?

ctapobep:
It's merged, but because I rebased the change GitHub doesn't recognize the commit as merged and thus I have to close it manually.

PR #62, opened 2015-03-16
JC-2130 Problem of crossing over buttons by white line is fixed

AnStrelnikov:
Here is explanation of problem:
http://stackoverflow.com/questions/12501224/
twitter-bootstrap-buttons-hover-style
Fixed by increasing background-position css style param


ctapobep:
@masyan could you please review the change?

PR #63, opened 2015-03-17
Sample implementation of component tests using Spock framework

mixas27:
Prepared test environment
Added sample test
Added sample classes for working with database


jcommune-view/jcommune-web-view/src/test/java/org/jtalks/jcommune/model/utils/Users.java
```diff
@@ -0,0 +1,96 @@
...
+        JCUser user = userDao.getByUsername(USERNAME);
+        if (user == null) {
+            user = new JCUser(USERNAME, "sample@example.com", encryptionService.encryptPassword(PASSWORD));
+            user.setEnabled(true);
+            user.addGroup(group);
+            userDao.saveOrUpdate(user);
+            userDao.flush();
+        }
+        //Needed for managing permissions
+        setAuthentication(new JCUser(USERNAME, "sample@example.com", PASSWORD));
```
ctapobep: That's why I'd call the method `signUpAndSignIn()` instead of `create()`.

Also, ideally every field of the user should be a random string (as it's done in system tests), but I think we'll do this afterwards.

mixas27:
merged

PR #65, opened 2015-03-19
JC-1669 Wrong page title when validation error happens during topic creation

wertklop:
Added changes for questions-n-answers-plugin

### DIFF
--- a/jcommune-plugins/questions-n-answers-plugin/src/main/resources/org/jtalks/jcommune/plugin/questionsandanswers/template/questionForm.vm
+++ b/jcommune-plugins/questions-n-answers-plugin/src/main/resources/org/jtalks/jcommune/plugin/questionsandanswers/template/questionForm.vm
@@ -19,7 +19,7 @@ Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
 <head>
   <meta name="description" content="${esc.html(${topicDto.topic.branch.name})}">
   <title>
-    #if(${topicDto.topic.branch.name})
+    #if(!${result.hasErrors()})
       ${esc.html(${topicDto.topic.branch.name})} -
     #end
     #if(${edit})


jcommune-plugins/questions-n-answers-plugin/src/main/resources/org/jtalks/jcommune/plugin/questionsandanswers/template/questionForm.vm
```diff
@@ -19,7 +19,7 @@ Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
 <head>
   <meta name="description" content="${esc.html(${topicDto.topic.branch.name})}">
   <title>
-    #if(${topicDto.topic.branch.name})
+    #if(!${result.hasErrors()})
```
mixas27: Why do we show name of branch in title only if validation errors happened? In this case if user will enter invalid data title of the page will be changed. I don't think that it is expected.

jcommune-plugins/questions-n-answers-plugin/src/main/java/org/jtalks/jcommune/plugin/questionsandanswers/controller/QuestionsAndAnswersController.java
```diff
@@ -148,7 +148,7 @@ public String createQuestion(@Valid @ModelAttribute TopicDto topicDto, BindingRe
         Map<String, Object> data = getDefaultModel(request);
         topicDto.getTopic().setBranch(branch);
         if (result.hasErrors()) {
-            data.put(BREADCRUMB_LIST, breadcrumbBuilder.getForumBreadcrumb(branch));
+            data.put(BREADCRUMB_LIST, breadcrumbBuilder.getForumBreadcrumb(topicDto.getTopic()));
```
mixas27: This fix problem for questions. It's OK. But what about discussions and code review? The problem exist for all topic types.

jcommune-plugins/questions-n-answers-plugin/src/main/resources/org/jtalks/jcommune/plugin/questionsandanswers/template/questionForm.vm
wertklop: It is expected, you can see description of issue

ctapobep:
Merged

PR #69, opened 2015-04-05
JC-2079 Fix topic modification date

NCNecros:
Add recalculation date after post delete

### DIFF
--- a/jcommune-model/src/main/java/org/jtalks/jcommune/model/entity/Topic.java
+++ b/jcommune-model/src/main/java/org/jtalks/jcommune/model/entity/Topic.java
@@ -189,8 +189,8 @@ public class Topic extends Entity implements SubscriptionAwareEntity {
      * @param post post to add
      */
     public void addPost(Post post) {
+        setModificationDate(post.getCreationDate());
         post.setTopic(this);
-        updateModificationDate();
         this.posts.add(post);
     }
 
@@ -201,6 +201,10 @@ public class Topic extends Entity implements SubscriptionAwareEntity {
      */
     public void removePost(Post postToRemove) {
         posts.remove(postToRemove);
+        Topic topic = postToRemove.getTopic();
+        if (postToRemove.getCreationDate().withMillis(0).equals(topic.getModificationDate().withMillis(0))) {
+            topic.recalculateModificationDate();
+        }
     }
 
     /**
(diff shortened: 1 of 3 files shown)


jcommune-model/src/main/java/org/jtalks/jcommune/model/entity/Topic.java
```diff
@@ -200,6 +200,7 @@ public void addPost(Post post) {
      * @param postToRemove post to remove
      */
     public void removePost(Post postToRemove) {
+        recalculateModificationDate();
```
mixas27: `recalculateModificationDate()` is called in `TransactionalPostService#deletePost` we should not duplicate this call. You should investigate why this method was not called in `TransactionalPostService#deletePost` and fix.

jcommune-model/src/test/java/org/jtalks/jcommune/model/entity/TopicTest.java
```diff
@@ -69,31 +69,25 @@ public void addPostShouldUpdateModificationDate() throws InterruptedException {
...
     @Test
-    public void updateModificationDateShouldChangeTheModificationDate() {
+    public void addPostShouldSetModificationDateToPostCreationDate() throws InterruptedException {
         Topic topic = createTopic();
         DateTime prevDate = topic.getModificationDate();
-
-        DateTime modDate = topic.updateModificationDate();
-
-        assertNotSame(modDate, prevDate);
+        Thread.sleep(25);
```
mixas27: I think, `Thread.sleep()` not a best choise in this case. You can set date manually and use methods like `date.plus Days()` to change it.

jcommune-model/src/test/java/org/jtalks/jcommune/model/entity/TopicTest.java
```diff
@@ -69,31 +69,25 @@ public void addPostShouldUpdateModificationDate() throws InterruptedException {
         Topic topic = createTopic();
         DateTime prevDate = topic.getModificationDate();
         Thread.sleep(25); // millisecond precise is a kind of fiction
-        topic.addPost(new Post());
+        Post post = new Post();
+        post.setCreationDate(new DateTime());
```
mixas27: Is it neccessary here? Same operation peformed in constructor.

jcommune-model/src/test/java/org/jtalks/jcommune/model/entity/TopicTest.java
```diff
@@ -192,6 +185,24 @@ public void removePostShouldRemovePostFromTheTopic() {
...
+        post1.setCreationDate(new DateTime());
+        Post post2 = new Post();
+        post2.setCreationDate(new DateTime());
+        Topic topic = new Topic(new JCUser(), "title");
+        topic.addPost(post1);
+        topic.addPost(post2);
+        Post post3 = new Post();
+        post3.setCreationDate(new DateTime());
+        topic.addPost(post3);
+        assertEquals(topic.getModificationDate(), post3.getCreationDate());
```
mixas27: This is case from another method. You should test only one case in test method.

ctapobep:
Merged!

PR #71, opened 2015-04-11
JC-2167 Incorrect English translation for delete question pop-up

Chichavl:
Renamed message constants.


ctapobep:
@Chichavl I see that you have commits for 2 tasks, could you please create 2 separate PRs?

PR #74, opened 2015-04-23
JC-2167 Incorrect English translation for delete question pop-up part2

Chichavl:
Renamed message constants.


ctapobep:
Thanks :)

PR #79, opened 2015-05-16

Pashokus:
Two validation messages properties files were changed. Only one letter was
changed from lowercase to uppercase.
Also unnecessary validation constrain was removed: -javax.validation.constraints.Size.message=Size must be between {min} and {max}		+title.length=Size must be between {min} - {max} characters
-#######################################################


ctapobep:
Merged!

PR #82, opened 2015-05-19

Pashokus:
Return usefull line "javax.validation.constraints.Size.message=Size must
be between {min} and {max}" of
D:\Java\Projects\jcommune\jcommune-view\jcommune-web-controller\src\main\resources\ValidationMessages_en.properties


ctapobep:
Merged

PR #83, opened 2015-05-19
JC-1269 issue fixed

artexnet:
- User can delete own starting post if no other user has added a reply, or those replies were removed/deleted by their owners.
- "Delete" button is also visible when there are several posts added only by initial topic starter.


ctapobep:
There is something wrong with this PR.. You've included a lot of changes in a lot of files (16).

artexnet:
Re-opened previous PR, going to close this one

PR #84, opened 2015-05-19
JC-1269 Allow deleting own topic if others didn't answer

artexnet:
- User can delete own starting post if no other user has added a reply, or those replies were removed/deleted by their owners
- "Delete" button is also visible when there are several posts added only by initial topic starter

### DIFF
--- a/jcommune-model/src/main/java/org/jtalks/jcommune/model/entity/Topic.java
+++ b/jcommune-model/src/main/java/org/jtalks/jcommune/model/entity/Topic.java
@@ -785,4 +785,18 @@ public class Topic extends Entity implements SubscriptionAwareEntity {
     public int getDisplayedPostsCount() {
         return getDisplayedPosts().size();
     }
+
+	/**
+	 * Checks if this topic contains only posts added by topic Owner.
+	 *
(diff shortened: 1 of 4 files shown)


ctapobep:
@mixas27 would you please do review one more time?

jcommune-model/src/main/java/org/jtalks/jcommune/model/entity/Topic.java
```diff
@@ -785,4 +785,18 @@ public Post getDraftForUser(JCUser user) {
     public int getDisplayedPostsCount() {
         return getDisplayedPosts().size();
     }
+
+	/**
+	 * Checks if this topic contains only posts added by topic Owner.
+	 *
+	 * @return <code>true</code> if condition is followed, otherwise <code>false</code>
+	 */
+	public boolean isContainsOwnerPostsOnly() {
```
mixas27: Don't forget about unit tests

artexnet:
Thanks @mixas27 , unit tests are added.

mixas27:
Please squash commits into single one and then I' merge

artexnet:
@mixas27 done.

mixas27:
Merged. Thank you.

PR #85, opened 2015-05-21
JC-2196 Change QnA answer button label to narrow down its meaning

Pashokus:
Change message for any registered users. Add message for question owner.
Add logic to detect who is owner and who is not, and named button based on
this.


ctapobep:
@Pashokus why the first line of the commit message is so weird (with underscores)?

Pashokus:
I copied name from name of branche. I will change it. Also I can use own variable. I will push commit with changes tomorrow

PR #86, opened 2015-05-22
JC-2196 Change QnA answer button label tonarrow down its meaning

Pashokus:
Change labelText for any registered users. Add labelText for question
owner. Add logic to detect who is owner and who is not, and named button based on this.


jcommune-plugins/questions-n-answers-plugin/src/main/resources/org/jtalks/jcommune/plugin/questionsandanswers/template/question.vm
```diff
@@ -107,6 +107,10 @@ Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
       #if(!${currentUser.anonymous} && ${post.userCreated.id} == ${currentUser.id})
         #set($own = true)
       #end
+      #set($labelAnswer = "label.answer")
+      #if($own)
```
mixas27: This condition is redundant.  We can determene which label to use without this flag.

mixas27:
Merged. Thank you.

PR #89, opened 2015-06-03

Pashokus:
remove useless annotation @NotBlank that generate second error "may not
empty"


mixas27:
Merged, thank you

PR #90, opened 2015-06-08
JC-2196 Change QnA answer button label tonarrow down its meaning

Pashokus:
add line that verify if user annonymus


ctapobep:
Merged

PR #92, opened 2015-06-08
JC-1501 Fix of the issue with links to users with '%'  in usernames

RuslanSh:
The problem arose because of two reasons: on the submit of post phase
there were decoding of the username which was not previously encoded.
It was solved by simply removing of that decoding in the
MentionedUsers.java. Also this fix demanded some changes in the
MentionedUsersTest.java. The problem on the preview phase arose because
of replacement of some special symbols. I assume that simple encoding is
enough and that's why I removed this replacement.


ctapobep:
Merged!

PR #95, opened 2015-06-19
JC-1930 Sign in functionality for not active users

RuslanSh:
There was added possibility to send email confirmation to the user,
who is trying to log in under not activated account. Authentication
status should not be true or false any more, because there can be
different conditions indicating different states of user's account.


jcommune-service/src/test/java/org/jtalks/jcommune/service/transactional/TransactionalUserServiceTest.java
```diff
@@ -503,11 +505,25 @@ public void testLoginUserWithBadCredentialsShouldFail()
...
-        boolean result = userService.loginUser(loginUserDto, httpRequest, httpResponse);
+        AuthenticationStatus result = userService.loginUser(loginUserDto, httpRequest, httpResponse);
 
-        assertFalse(result, "Login user with bad credentials should fail.");
+        assertEquals(result, AuthenticationStatus.AUTHENTICATION_FAIL, "Login user with bad credentials should fail.");
+    }
+
+    @Test
+    public void testLoginNotActivatedUserShouldFail()
+            throws UnexpectedErrorException, NoConnectionException {
```
ctapobep: It doesn't matter in tests what exceptions get thrown, so it's easier to `throws Exception`.

jcommune-service/src/test/java/org/jtalks/jcommune/service/transactional/TransactionalAuthenticatorTest.java
```diff
@@ -198,11 +201,32 @@ public void authenticateUserWithNewCredentialsShouldBeSuccessful() throws Except
...
+                "Authentication user with new credentials should be successful.");
+    }
+
+    @Test
+    public void authenticateNotEnabledUserShouldFail() throws Exception {
+            String password = "password";
+            String passwordHash = "5f4dcc3b5aa765d61d8327deb882cf99";
+            LoginUserDto loginUserDto = createDefaultLoginUserDto();
+            JCUser oldUser = prepareOldUser(loginUserDto.getUserName());
+            when(userDao.getByUsername(oldUser.getUsername())).thenReturn(oldUser);
```
ctapobep: Do you think maybe there is a way shortening the test method and removing extra when()?

RuslanSh: Thank you, both your notes make sense and I've fixed it.

ctapobep:
Trying to login as a non-activated user, but leave password field empty: 
```We're sorry! The server encountered an internal error or misconfiguration and was unable to complete your request.``

RuslanSh:
@ctapobep Ok, I'll work on fixing it. Which behavior is expected in this case?

ctapobep:
Same behavior as before the activation stuff was brought up.

PR #98, opened 2015-07-01
JC-1739 Fix of issue with repeated post request after refresh

RuslanSh:
There was changed error handling behavior. Now, instead of redirecting
to view it redirects to an error handling controller method. Thus, when
user refreshes page only GET request is repeated. In consequence, it
allows saving drafts and prevents unexpected text body errors.

### DIFF
--- a/jcommune-view/jcommune-web-controller/src/main/java/org/jtalks/jcommune/web/controller/PostController.java
+++ b/jcommune-view/jcommune-web-controller/src/main/java/org/jtalks/jcommune/web/controller/PostController.java
@@ -41,6 +41,7 @@ import org.springframework.validation.BindingResult;
 import org.springframework.web.bind.WebDataBinder;
 import org.springframework.web.bind.annotation.*;
 import org.springframework.web.servlet.ModelAndView;
+import org.springframework.web.servlet.mvc.support.RedirectAttributes;
 import org.springframework.web.util.WebUtils;
 
 import javax.servlet.http.HttpServletRequest;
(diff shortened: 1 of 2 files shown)


jcommune-view/jcommune-web-controller/src/main/java/org/jtalks/jcommune/web/controller/PostController.java
```diff
@@ -253,6 +243,30 @@ public ModelAndView create(@RequestParam(value = "page", defaultValue = "1", req
         return new ModelAndView(this.redirectToPageWithPost(newbie.getId()));
     }
 
+    @RequestMapping(method = RequestMethod.GET, value = "/topics/error/{topicId}")
+    public ModelAndView errorRedirect(@RequestParam(value = "page", required = false) String page,
```
ctapobep: Could you please add javadocs explaining why this is important?

ctapobep:
@standpoint would you please review the change?

jcommune-view/jcommune-web-controller/src/test/java/org/jtalks/jcommune/web/controller/PostControllerTest.java
```diff
@@ -223,13 +224,13 @@ public void testSubmitAnswerValidationFail() throws NotFoundException {
...
         //invoke the object under test
-        ModelAndView mav = controller.create(null, TOPIC_ID, getDto(), resultWithErrors);
+        ModelAndView mav = controller.create(null, TOPIC_ID, getDto(), resultWithErrors, new RedirectAttributesModelMap());
 
         //check expectations
         verify(topicModificationService, never()).replyToTopic(anyLong(), anyString(), eq(BRANCH_ID));
 
         //check result
-        assertViewName(mav, "topic/postList");
+        assertEquals(mav.getViewName(), "redirect:/topics/error/" + TOPIC_ID + "?page=null");
```
standpoint: @RuslanSh , before fix the test was check assertion about redirection to the topic view page. Now this one check another behaviour - redirection when validation result had errors.
Perhap it would be better to check these two distinct behaviours in two separated test-methods?

standpoint:
In general, all looks quite acceptable for me (besides unit-test)
In addition, according to last issue, there is needs to fix the same thing for the Questions&Answears

jcommune-view/jcommune-web-controller/src/test/java/org/jtalks/jcommune/web/controller/PostControllerTest.java
RuslanSh: @standpoint , this test is called testSubmitAnswerValidationFail(), thus it checks redirect after validation fail as well as it did before. The only difference is that previously it was redirect to the view and now it is redirect to another action. Do you think I have to write any extra tests? If so, please, let me know what you want me to check.

RuslanSh:
@standpoint I think you are right, but I am not sure that we have to do it in the scope of this backlog. Do we have a separate backlog for the Q&A?

jcommune-view/jcommune-web-controller/src/test/java/org/jtalks/jcommune/web/controller/PostControllerTest.java
standpoint: @RuslanSh you are right! I'm sorry, I must be looked at the previews test name "testSubmitAnswerValidationPass"
so, I have no more remarks to this one.

standpoint:
@ctapobep if there are no more comments, could you please merge the fix?

PR #100, opened 2015-07-13
Poulpe-528 Issue with 'User #0 not found' error

standpoint:
We have to customize serialization of the entity JCUser because of we used to loose 'id' and 'uuid' data from org.jtalks.common.model.entity.Entity class when saved other User details in session-storage-file.
The fix is similar to issue of PoulpeUser serialization (http://jira.jtalks.org/browse/POULPE-528)

Created:
  - methods writeObject(), readObject() and serialVersionUID field.
  - test method for checking is most fields of User entity serialized
  - test method to assert the field `List<Group> groups` can't be serialized with rest of the JCUser fields.

Implemented additional ObjectFactory method to reduce size of code in test method.

### DIFF
--- a/jcommune-model/src/test/java/org/jtalks/jcommune/model/entity/JCUserTest.java
+++ b/jcommune-model/src/test/java/org/jtalks/jcommune/model/entity/JCUserTest.java
@@ -20,8 +20,10 @@ import org.jtalks.common.model.entity.Group;
 import org.jtalks.common.model.entity.User;
 import org.springframework.security.core.GrantedAuthority;
 import org.springframework.security.core.authority.GrantedAuthorityImpl;
+import org.springframework.util.SerializationUtils;
 import org.testng.annotations.Test;
 
+import java.lang.reflect.InvocationTargetException;
 import java.util.Arrays;
 import java.util.List;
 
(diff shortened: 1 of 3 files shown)


jcommune-model/src/test/java/org/jtalks/jcommune/model/entity/JCUserTest.java
```diff
@@ -217,6 +220,37 @@ public void spanishLanguageHandling () {
         assertEquals(language, Language.ENGLISH);
     }
 
+    @Test
+    public void entityFieldsShouldBeSerialized() throws NoSuchMethodException, InvocationTargetException,
+            IllegalAccessException {
+        JCUser user = ObjectsFactory.getUserWithAllNotInheritedFieldsFilled();
+        user.setAvatar(new byte[]{1});
```
ctapobep: So why don't use fill all these fields in the ObjectsFactory?

standpoint: I reasoned thus these fields are the part of `jtalks-common` API and wondered would it be really right approach to set these ones at the jcommune' `ObjectFactory`, (moreover there are needs to use reflection to set up some fields)?
If this is ok I'll fill all these fields at `ObjectFactory`

ctapobep: We still fill them even if we fill them in tests :) I prefer to move them to `ObjectsFactory` to keep the tests as readable as possible.

ctapobep:
@mixas27 would you please review the changes? Similar changes were introduced in Poulpe recently.

mixas27:
Looks good in my opinion

PR #104, opened 2015-07-28
Modify sql-script

virtus80:
Fixing problem with conflicting groups which were created via poulpe and sql-script.


ctapobep:
@virtus80 please be more specific what was the change and what is the intention of it.

PR #105, opened 2015-07-31
JC-1801. Add domain жж.рф to livejournal validation contact pattern

hacke151:
JC-1801. Add domain жж.рф to livejournal validation contact pattern


ctapobep:
Hello there, we thought about this and decided that we need to [get rid of validation in contacts|http://jira.jtalks.org/browse/JC-2234] at all.
But anyway the suggested code is not correct :) To change anything in DB schema you'd need to create additional SQL migration instead of updating existing ones. We use Flyway which won't apply the same migration to DB if it was applied before.

PR #106, opened 2015-08-01
Modify sql-script JC-2145

virtus80:
Added 'ignore' in insertion into acl_object_identity to prevent
re-creared groups that were created via poulpe.
Added new variables for insertion personal permissions in acl_entry. If
groups created only via script new variables is equals to old ones (
@..._group_object_identity), but it prevents from link to unexisting id
when groups were created via poulpe.


ctapobep:
@standpoint could you please review the changes?

standpoint:
In my opinion, changes looks good. And works also well.

PR #108, opened 2015-08-15
JC-1178 Display of hyperlink in topics/posts body

Vitalij-Voronkoff:
The script finds matches with regular expression string, adds <a href> tag
and replace string in a html page

Next Step: The rules for external/internal links should be yhe same as for
[url] bb code. External links should be prefixed with /out?url= part and
have no follow attribute.


ctapobep:
Fixed in another PR

PR #115, opened 2015-09-04
Create branch component test

mixas27:
Implemented base for component tests of branch creation. Also test code was restructured


jcommune-view/jcommune-web-view/src/test/java/org/jtalks/jcommune/test/CreateBranchTest.groovy
```diff
@@ -0,0 +1,158 @@
...
+    private BranchService branchService
+    @Autowired
+    private GroupsService groupsService
+
+    private MockMvc mockMvc
+
+    private Component forum
+
+    def setup() {
+        mockMvc = MockMvcBuilders.webAppContextSetup(ctx)
```
ctapobep: Let's crete this mockMvc in Spring context and set it into `branches` there (in Spring).

jcommune-view/jcommune-web-view/src/test/java/org/jtalks/jcommune/test/CreateBranchTest.groovy
```diff
@@ -0,0 +1,158 @@
...
+    private WebApplicationContext ctx
+    @Resource(name = 'testFilters')
+    List<Filter> filters
+
+    @Autowired
+    private Branches branches;
+    @Autowired
+    private ComponentService componentService
+    @Autowired
+    private UserService userService
```
ctapobep: Hm.. I just noticed that there are 2 different classes: `UserService` and `Users`. In these tests we should use only `Users` - we don't want to have confusion between these.

jcommune-view/jcommune-web-view/src/test/java/org/jtalks/jcommune/test/CreateBranchTest.groovy
```diff
@@ -0,0 +1,158 @@
...
+                .addFilters(filters.toArray(new Filter[filters.size()])).build()
+        branches.mockMvc = mockMvc
+        groupsService.create()
+        forum = componentService.createForumComponent()
+    }
+
+    def 'test create branch success'() {
+        given: 'User created and have admin permission on forum'
+            def user = new User()
+            userService.create(user).withPermissionOn(forum, GeneralPermission.ADMIN)
```
ctapobep: If we use `Users` instead of `UserService` this can become: `users.signUpForumAdmin()` or:
```java
User user = users.signUp()
users.grantAdminPermissions(user)
```

jcommune-view/jcommune-web-view/src/test/java/org/jtalks/jcommune/test/CreateBranchTest.groovy
```diff
@@ -0,0 +1,158 @@
...
+    List<Filter> filters
+
+    @Autowired
+    private Branches branches;
+    @Autowired
+    private ComponentService componentService
+    @Autowired
+    private UserService userService
+    @Autowired
+    private BranchService branchService
```
ctapobep: We should work only with `Branches` here.

jcommune-view/jcommune-web-view/src/test/java/org/jtalks/jcommune/test/CreateBranchTest.groovy
```diff
@@ -0,0 +1,158 @@
...
+
+    @Autowired
+    private WebApplicationContext ctx
+    @Resource(name = 'testFilters')
+    List<Filter> filters
+
+    @Autowired
+    private Branches branches;
+    @Autowired
+    private ComponentService componentService
```
ctapobep: Would be great to remove it from the test layer and leave it somewhere below..

jcommune-view/jcommune-web-view/src/test/java/org/jtalks/jcommune/test/CreateBranchTest.groovy
```diff
@@ -0,0 +1,158 @@
...
+            def user = new User()
+            userService.create(user).withPermissionOn(forum, GeneralPermission.ADMIN)
+        and: 'User signed in'
+            def session = userService.signIn(mockMvc, user)
+        when: 'User creates branch'
+            def branch = new Branch(name: branchName)
+            branches.create(branch, session)
+        then: 'Validetion error occurs'
+            def e = thrown(ValidationException)
+            [errorMessage].equals(e.defaultErrorMessages)
```
ctapobep: equals() can be replaced with ==

jcommune-view/jcommune-web-view/src/test/java/org/jtalks/jcommune/test/CreateBranchTest.groovy
```diff
@@ -0,0 +1,158 @@
...
+            !branchService.isExist(branch.name)
+        where:
+        description             |errorMessage                                               |caseName
+        randomAlphabetic(256)   |'Branch description length should be 255 characters max'   |'Branch description length greather than 255 characters'
+    }
+
+    def 'create branch should fail if user have no permissions'() {
+        given: 'User created and have no admin permission on forum'
+            def user = new User()
+        and: 'User signed in'
```
ctapobep: But where did we sign it up?

jcommune-view/jcommune-web-view/src/test/java/org/jtalks/jcommune/test/service/UserService.groovy
```diff
@@ -0,0 +1,134 @@
...
+        def token = new UsernamePasswordAuthenticationToken(user.username, user.password)
+        token.details = user
+        def auth = authenticationManager.authenticate(token)
+        securityFacade.context.authentication = auth
+        def request = new MockHttpServletRequest()
+        def response =  new MockHttpServletResponse()
+        sessionStrategy.onAuthentication(auth, request, response)
+    }
+
+
```
ctapobep: Too many spaces :)

ctapobep:
It became unclear what's the different between `UserService` and `Users`. I'd suggest to have only `Users`, `Branches`, etc which would contain all the necessary business functions.

mixas27:
My idea is to create 2 separate classes for each entity, for example `Users` and `UserService`. So one of them will be used for test proposes (perform operations using http request and asserting results) another for performing operations without making http requests (by calling DAO's methods) and performing assertions.

mixas27:
I think it help to distinguish check of test cases from providing preconditions.

ctapobep:
Why do tests need to know the difference? If we have something implemented via "HTTP", then it should always go there. 

It's just the purpose of making this layer is to make it readable for non-technical people (if needed). And this new concept may complicate understanding.

jcommune-view/jcommune-web-view/src/test/java/org/jtalks/jcommune/test/CreateBranchTest.groovy
ctapobep: This should work as Factory Bean that creates the `MockMvc`:
```groovy
import org.springframework.beans.factory.FactoryBean
import org.springframework.beans.factory.annotation.Autowired
import org.springframework.test.web.servlet.MockMvc
import org.springframework.test.web.servlet.setup.MockMvcBuilders
import org.springframework.web.context.WebApplicationContext

class MockMvcFactoryBean implements FactoryBean<MockMvc> {
    @Autowired WebApplicationContext applicationContext

    @Override
    MockMvc getObject() throws Exception {
        return MockMvcBuilders.webAppContextSetup(applicationContext).build()
    }

    @Override
    Class<?> getObjectType() {
        return MockMvc
    }

    @Override
    boolean isSingleton() {
        return true
    }
}
```

ctapobep:
Merged

PR #118, opened 2015-09-11
JC-2234.Remove validation from contacts

Vitalij-Voronkoff:
-When choosing a particular contact type it should be able to put anything
into the input field and no validation will happen.
-The only constraint should be 255 symbols max.
-In case contact is too large when looking at it on the UI, it should end
with "..." and the rest of the contact should not be shown.But it still
should be clickable.


ctapobep:
@Vitalij-Voronkoff shouldn't we also change SQL scripts? I remember hearing this from you..

mixas27:
In my opinion, annotation `@ValidUserContact` and appropriate validator should be removed. This validator is too specific and I not see any other cases to use this validator. Also I'd recommend to remove `validationPattern` field from `UserContactType` since we don't need this information. 
PS: don't forget to remove unused imports from classes

mixas27:
Merged

PR #123, opened 2015-09-29
JC-2241 Internal server error in RSS page 

despc:
Fix RSS content with invalid XML chars in post body

### DIFF
--- a/jcommune-model/src/main/java/org/jtalks/jcommune/model/entity/ObjectsFactory.java
+++ b/jcommune-model/src/main/java/org/jtalks/jcommune/model/entity/ObjectsFactory.java
@@ -16,7 +16,9 @@ package org.jtalks.jcommune.model.entity;
 
 import org.apache.commons.lang.math.RandomUtils;
 import org.joda.time.DateTime;
-import org.jtalks.common.model.entity.*;
+import org.jtalks.common.model.entity.Group;
+import org.jtalks.common.model.entity.Section;
+import org.jtalks.common.model.entity.User;
 
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
@@ -253,5 +255,4 @@ public final class ObjectsFactory {
         post.addComment(new PostComment());
         return post;
     }
-
 }
--- a/jcommune-view/jcommune-web-controller/src/main/java/org/jtalks/jcommune/web/util/RssUtils.java
+++ b/jcommune-view/jcommune-web-controller/src/main/java/org/jtalks/jcommune/web/util/RssUtils.java
@@ -1,0 +1,41 @@
+/**
+ * Copyright (C) 2011  JTalks.org Team
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ */
+package org.jtalks.jcommune.web.util;
+
+import org.apache.commons.lang.StringUtils;
+
+/**
+ * @author denis.berezhnoy
+ */
+public class RssUtils {
+    /**
+     * Method for skip invalid characters in content for RSS feed.
+     * <p/>
+     * Description valid chars in XML specification
+     * http://www.w3.org/TR/REC-xml/#charsets
+     *
+     * @param in - post content.
+     */
+    public static String skipInValidXMLChars(String in) {
+        if (StringUtils.isBlank(in)) return StringUtils.EMPTY;
+        String pattern = "[^"
+                + "\u0009\r\n"
+                + "\u0020-\uD7FF"
+                + "\uE000-\uFFFD"
+                + "\ud800\udc00-\udbff\udfff"
+                + "]";
+        return in.replaceAll(pattern, StringUtils.EMPTY);
+    }
+}
--- a/jcommune-view/jcommune-web-controller/src/test/java/org/jtalks/jcommune/web/view/RssViewerTest.java
+++ b/jcommune-view/jcommune-web-controller/src/test/java/org/jtalks/jcommune/web/view/RssViewerTest.java
@@ -15,8 +15,8 @@
 
 package org.jtalks.jcommune.web.view;
 
-import com.sun.syndication.feed.rss.Channel;
 import com.sun.syndication.feed.rss.Item;
+import com.sun.syndication.feed.rss.Channel;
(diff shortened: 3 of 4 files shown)


jcommune-view/jcommune-web-controller/src/main/java/org/jtalks/jcommune/web/util/RSSUtils.java
```diff
@@ -0,0 +1,43 @@
...
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ */
+package org.jtalks.jcommune.web.util;
+
+/**
+ * @author denis.berezhnoy
+ */
+public class RSSUtils {
```
ctapobep: Should be `RssUtils`

jcommune-view/jcommune-web-controller/src/main/java/org/jtalks/jcommune/web/util/RSSUtils.java
```diff
@@ -0,0 +1,43 @@
...
+     * @param in - post content.
+     */
+    public static String skipInValidXMLChars(String in) {
+        StringBuffer out = new StringBuffer();
+        char current;
+
+        if (in == null || ("".equals(in))) return "";
+        for (int i = 0; i < in.length(); i++) {
+            current = in.charAt(i);
+            if ((current == 0x9) ||
```
ctapobep: These should be constants. Also you should explain why these chars are invalid for XML. And it seems to me that you're just skipping the characters.. That's an important decision that should be carefully documented and justified.

jcommune-view/jcommune-web-controller/src/main/java/org/jtalks/jcommune/web/util/RSSUtils.java
```diff
@@ -0,0 +1,43 @@
...
+    /**
+     * Method for skip invalid characters in content for RSS feed.
+     *
+     * @param in - post content.
+     */
+    public static String skipInValidXMLChars(String in) {
+        StringBuffer out = new StringBuffer();
+        char current;
+
+        if (in == null || ("".equals(in))) return "";
```
ctapobep: isBlank() in apache commons

jcommune-view/jcommune-web-controller/src/test/java/org/jtalks/jcommune/web/view/RssViewerTest.java
```diff
@@ -69,18 +72,32 @@ protected void setUp() {
         topics.add(topic);
         topics.add(topic);
         model.put("topics", topics);
+        List<Topic> topicsWithSpecChars = new ArrayList<>();
```
ctapobep: All this code needs to be moved into the test, it doesn't belong to the precondition. You can encapsulate some of it into private methods or use ObjectsFactory to keep the test clean.

jcommune-view/jcommune-web-controller/src/test/java/org/jtalks/jcommune/web/view/RssViewerTest.java
```diff
@@ -69,18 +72,32 @@ protected void setUp() {
...
     public void testBuildFeedItems() throws Exception {
 
         List<Item> items = rssViewer.buildFeedItems(model, request, response);
         assertEquals(items.get(0).getAuthor(), "username");
-        assertEquals(items.get(0).getComments(), "Signature");
         assertEquals(response.getContentType(), rssViewer.getContentType());
     }
 
     @Test
+    public void testBuildFeedItemsWithPostHaveSpecChars() throws Exception {
```
ctapobep: Test names should describe behavior. E.g. `feedBodyMustStripSpecialSymbols`

jcommune-view/jcommune-web-controller/src/test/java/org/jtalks/jcommune/web/view/RssViewerTest.java
```diff
@@ -120,4 +137,18 @@ public void testRssFeed() throws Exception {
         verify(rssViewerMock).buildFeedItems(model, request, response);
     }
 
+    private boolean containtsOnlyValidXMLChars(String stringToValidate) {
+        char current;
+        for (int i = 0; i < stringToValidate.length(); i++) {
+            current = stringToValidate.charAt(i);
+            if (!((current == 0x9) ||
```
ctapobep: Why not refer to the symbols by their String value?

jcommune-view/jcommune-web-controller/src/test/java/org/jtalks/jcommune/web/view/RssViewerTest.java
```diff
@@ -120,4 +109,17 @@ public void testRssFeed() throws Exception {
         verify(rssViewerMock).buildFeedItems(model, request, response);
     }
 
+    private boolean containtsOnlyValidXMLChars(String stringToValidate) {
+        String pattern = "[^"
```
ctapobep: Why not simply iterate over the array of chars and not check if `stringToValidate.contains(character)`.

jcommune-model/src/main/java/org/jtalks/jcommune/model/entity/ObjectsFactory.java
```diff
@@ -254,4 +256,34 @@ public static Post getPostWithComments() {
         return post;
     }
 
+    public static Map<String, Object>  getNewsModelWithSpecChars(){
```
ctapobep: Model? This method should create a post, a topic or something else. But there is no such entity as model.. Makes sense to use `getDefaultTopic()` or something like that where you replace the body in the test. Let's have a talk in the evening if you need further explanation.

jcommune-view/jcommune-web-controller/src/main/java/org/jtalks/jcommune/web/util/RssUtils.java
```diff
@@ -0,0 +1,41 @@
...
+     * <p/>
+     * Description valid chars in XML specification
+     * http://www.w3.org/TR/REC-xml/#charsets
+     *
+     * @param in - post content.
+     */
+    public static String skipInValidXMLChars(String in) {
+        if (StringUtils.isBlank(in)) return StringUtils.EMPTY;
+        String pattern = "[^"
+                + "\u0009\r\n"
```
ctapobep: Could you please check the difference in performance between regex and replaceAll for individual characters?

jcommune-model/src/main/java/org/jtalks/jcommune/model/entity/ObjectsFactory.java
```diff
@@ -254,4 +256,32 @@ public static Post getPostWithComments() {
...
+        Post postWithSpecChars = new Post(user, "����\u000F���");
+        Topic topicWithSpecChars = new Topic(user, "");
+        topicWithSpecChars.addPost(postWithSpecChars);
+        topicWithSpecChars.setId(2L);
+        topicsWithSpecChars.add(topicWithSpecChars);
+
+        return topicsWithSpecChars;
+    }
+
+    public static Map<String, Object>  getNewsModel()  {
```
ctapobep: Hm.. I was sure I wrote this.. There is no such thing in the system as NewsModel. Therefore people won't understand it. You need to return a list of topics, not a map of abstract things.

ctapobep:
Could you please fix the last comment and then squash all commits into 1?

PR #128, opened 2015-10-05
JC-2126 On Sign In popup window cursor is not focused on link

vps2:
- fixed tabNavigation property of the Sign-In dialogue.

P/S: note at the bottom of this issue: "cursor is not focused on “Save” button when moving via Tab key" is not confirmed, because I have the focus on Save button via Tab (in browsers Opera 12.17, Crome 45.0.2454.101 m).


ctapobep:
Thanks :)

PR #129, opened 2015-10-06
JC-1998 Wrong error message below the "New Password" field

vps2:
- added a message attribute of the annotation Size (the newUserPassword
  field of the UserSecurityDto class)

### DIFF
--- a/jcommune-view/jcommune-web-controller/src/main/java/org/jtalks/jcommune/web/dto/UserSecurityDto.java
+++ b/jcommune-view/jcommune-web-controller/src/main/java/org/jtalks/jcommune/web/dto/UserSecurityDto.java
@@ -36,7 +36,8 @@ public class UserSecurityDto {
     private long userId;
 
     private String currentUserPassword;
-    @Size(min = User.PASSWORD_MIN_LENGTH, max = User.PASSWORD_MAX_LENGTH)
+    @Size(min = User.PASSWORD_MIN_LENGTH, max = User.PASSWORD_MAX_LENGTH,
+            message = "{user.password.length_constraint_violation}")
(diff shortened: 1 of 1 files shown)


ctapobep:
No need to put underscores in the commit message - it's not a branch or a tag name :)

vps2:
Ok :)

ctapobep:
Reviewed - the solution doesn't comply with the expected result. Instead of using custom labels for every field we want to use a single label for all of the fields. What we need to do is to change the label to the common one (the one that all the other fields are using) and then remove `user.password.length_constraint_violation` at all from all the files.

vps2:
In the resource bundle "ValidationMessages" has label  "length.constraint" with the following content:

- eng -> "Length must be between {min} and {max} symbols"
- rus -> "Длина должна быть между 1 и 50 символами"
- urk -> "Довжина повинна бути від 1 до 50 символів"

So, we get the expected result. This is the right label that will be used by all other fields?

ctapobep:
Yap

vps2:
Sorry, but I can't understand You.
Can i something not so wrote?

ctapobep:
You should use the generic message, `length.constraint` seems to be the right one.

vps2:
It is now clear. 
Thank you.

vps2:
You probably messed up the second letter in the word above (https://github.com/jtalks-org/jcommune/pull/129#issuecomment-145966736). According to the dictionary, this word a different meaning.

ctapobep:
They can use any letter there: yap, yup, yep, etc. In English there are no boundaries :)

vps2:
Well then I misunderstood. :)

PR #131, opened 2015-10-07
JC-2250 Error messages are not shown for anonymous user when he tries to vote up/down

vps2:
- added check for anonymous user when inserting the tooltip for
  vote buttons.


ctapobep:
Merged, thanks a lot!

PR #132, opened 2015-10-08
JC-2092 There is no confirmation popup when leaving pages with unsaved data

vps2:
- banned the forwarding of the "click" event upwards to parent elements.

P/S: according to the jquery documentation, any event which occurs on a child element propagate up the hierarchy. So when you click on any element occurs unsubscribing from the events of "beforeunload". To prevent this, you should use "event.stopPropagation()". But there's a catch - this method does not work if the subscription to the event (for example "click") occurs within the method "$('element').on()". Because of this I had to use a loop "$.each()" to the array elements.


ctapobep:
Ouch.. I think this issue became obsolete since we're implementing Topic Drafts (the body will be saved as we type), so there won't be any risks of loosing the data.

Sorry for that!

vps2:
ok :)

PR #134, opened 2015-10-11
JC-2071 Additional incorrect tooltip over the button "Upload" in User's "Profile" page

vps2:
- fixed the style and behavior of "input[type='file']".
- added an event handler "onCkick" to the element "a#upload".
- overriden method of the upload button.

### DIFF
--- a/jcommune-view/jcommune-web-view/src/main/webapp/resources/javascript/app/avatarUpload.js
+++ b/jcommune-view/jcommune-web-view/src/main/webapp/resources/javascript/app/avatarUpload.js
@@ -70,8 +70,10 @@ $(document).ready(function () {
         action = $root + '/users/XHRavatarpreview';
     }
 
+    var fictiveButton = $("#upload").get(0);
+
     var uploader = new qq.FileUploaderBasic({
-        button: $("#upload").get(0),
+        button: fictiveButton,
         //server side uploading handler
         action: action,
         //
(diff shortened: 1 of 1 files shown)


jcommune-view/jcommune-web-view/src/main/webapp/resources/javascript/app/avatarUpload.js
```diff
@@ -70,8 +70,10 @@ $(document).ready(function () {
         action = $root + '/users/XHRavatarpreview';
     }
 
+    var fictiveButton = $("#upload").get(0);
```
ctapobep: Fictive? Why is it fictive?

ctapobep:
In the commit message I see "what", but it's not clear "why" and "why not something else"..

vps2:
Fictive - because the tag to which it refers is only a wrapper of input element. Can offer its name?
Can I write the answer in russian to a question about comments in the commit (need write a lot)?

vps2:
For the text of tooltips in the item "input[type=file]" meets the "title" attribute. If it is empty, the tooltip does not appear. But there is a problem is that in different browsers under empty means different (in some it is "", and in others it is " ").
Thus, once the tooltip can be removed in two ways: 
1. replace the "title" attribute of element "input.
2. move element "input[type=file]" beyond the wrapper element (<a id='upload'>) to the hover event of the mouse cursor did not begin with the element "input". Further, the event "click" of the element "input" to trigger using the event "click" of the wrapper element and its contained elements (this way i went).

ctapobep:
That sounds reasonable :) Could you please put this info into commit message and/or comments in the code?

vps2:
Did...

ctapobep:
Cool, merged 8-)

PR #137, opened 2015-10-14
JC-2198 Rename "Save" Button on Topic and CR creation page

SurfVaporizer:
- All initial buttons to create topics should be: Send
- All editing buttons (both for post and topic) should be: Edit
- All the buttons to save corrections should be: Save
- Buttons to initially create post should stay unchanged


ctapobep:
@SurfVaporizer could you please squash these 2 commits into 1?

SurfVaporizer:
Squashed.

ctapobep:
Merged

PR #138, opened 2015-10-14
JC-2086 It's impossible to refresh CAPTCHA via Tab

SurfVaporizer:
- Can't move focus to "Captcha text" field via Tab in Firefox - fixed

### DIFF
--- a/jcommune-view/jcommune-web-view/src/main/webapp/WEB-INF/jsp/registration.jsp
+++ b/jcommune-view/jcommune-web-view/src/main/webapp/WEB-INF/jsp/registration.jsp
@@ -32,7 +32,7 @@
 
 <div class="container form-login-related registration-page">
   <form:form id="form" name="form" action='${pageContext.request.contextPath}/user/new'
-             modelAttribute="newUser" method="POST" class='form-vertical'>
+             modelAttribute="newUser" method="POST" class='form-vertical' tabindex="-1">
     <fieldset>
       <legend><spring:message code="label.fillmessage"/></legend>
       <div class="control-group">
--- a/jcommune-view/jcommune-web-view/src/main/webapp/resources/javascript/app/registration.js
+++ b/jcommune-view/jcommune-web-view/src/main/webapp/resources/javascript/app/registration.js
@@ -33,7 +33,6 @@ $(function () {
         captchaContainer.find('.captcha-img, .btn-captcha-refresh').on({
             click: refreshCaptchaJsp,
             keypress: function(e) {
-                e.preventDefault();
                 if (isEnterKeyPressed(e)) {
                     refreshCaptchaJsp();
                 }


jcommune-view/jcommune-web-view/src/main/webapp/WEB-INF/jsp/registration.jsp
```diff
@@ -32,7 +32,7 @@
 
 <div class="container form-login-related registration-page">
   <form:form id="form" name="form" action='${pageContext.request.contextPath}/user/new'
-             modelAttribute="newUser" method="POST" class='form-vertical'>
+             modelAttribute="newUser" method="POST" class='form-vertical' tabindex="-1">
```
masyan: зачем это тут? у нас порядок фокуса по табу определяется в js генерирующем диалоговое окно

SurfVaporizer: на самом деле можно убрать, я не мог разобраться почему не фокус по табу слетает в FF.

ctapobep: So AFAIU the focus order has issues in FF, @masyan maybe you could take a look and help @SurfVaporizer with that? 
@SurfVaporizer did I understand the situation correctly?

SurfVaporizer: @ctapobep now the focus issue is solved in FF. Focus work correctly with tabindex param in the form or without it.

ctapobep: So, @masyan would you be able to have a look why our dialog JS doesn't work for FF in this case?

masyan: it works. on my ff 41.0.2/linux/preprod

ctapobep: Try on QA, PREPROD has an old version. Make sure you're trying the page, not the dialog.

masyan:
works fine. (QA)

jcommune-view/jcommune-web-view/src/main/webapp/WEB-INF/jsp/registration.jsp
masyan: works fine. (QA)

ctapobep: That's very weird.. I can reproduce the issue on QA with my FF 41.0.2. Maybe there is a difference between FFs on different platform though that's still weird. 

So given the bug is reproduced on others' machines, do you think that current fix is applicable?

ctapobep:
This has already been merged, right?

PR #139, opened 2015-10-14
JC-2259 Username which consists of [] does not become a link.

despc:


jcommune-service/src/main/java/org/jtalks/jcommune/service/nontransactional/BBCodeService.java
```diff
@@ -73,13 +74,21 @@ public String convertBbToHtml(String bbEncodedText) {
         for (TextProcessor preprocessor : preprocessors) {
             bbEncodedText = preprocessor.process(bbEncodedText);
         }
-        bbEncodedText = processor.process(bbEncodedText);
+        if (isUserbbCode(bbEncodedText)){
```
ctapobep: I'm not sure I fully understand the changes - why do we process text differently depending on whether `[user]` bb code is present?

jcommune-service/src/test/java/org/jtalks/jcommune/service/nontransactional/BBCodeServiceTest.java
```diff
@@ -41,6 +48,15 @@ public void testQuote() {
         assertEquals(result, "[quote=\"name\"]source[/quote]");
     }
 
+    @Test
+    public void testUserbbProcessorReturnCorrectLink() {
```
ctapobep: I think you mean nested `[user]` tags here? I'd update the test name to be more precise.

jcommune-service/src/main/java/org/jtalks/jcommune/service/nontransactional/BBCodeService.java
despc: In case when bb code contains tag [user], text processing execute with help another algoritm because kefirrbb does not work properly with eg "[user=/jcommune/users/16][user]user[/user][/user]" i.e user name contains tag [user]

ctapobep: Could you please add comments with the details? In general all non-typical solutions should have solid comments.

ctapobep:
Ok, could you squash the commits into 1 and I'll merge them?

ctapobep:
Merged!

PR #141, opened 2015-10-17
JC-2115 Captcha text field has different length on pop-up window and on registration page

vps2:
- in the HTML representation of the plugin removed the class that provides
  the size of the input field.
- in the script that creates a registration form added code that insert
  style to the captcha input field (as in the rest form inputs).
- fixed tests "testGetHtml()" of the plugin.


ctapobep:
Merged, thanks!

PR #142, opened 2015-10-19

dsdolzhenko:
- add client side code for draft topic
 - change constraints for some fields of TopicDraft entity

### DIFF
--- a/jcommune-model/src/main/java/org/jtalks/jcommune/model/entity/TopicDraft.java
+++ b/jcommune-model/src/main/java/org/jtalks/jcommune/model/entity/TopicDraft.java
@@ -19,7 +19,6 @@ import org.jtalks.common.model.entity.Entity;
 import org.jtalks.jcommune.model.validation.annotations.AtLeastOneFieldIsNotNull;
 import org.jtalks.jcommune.model.validation.annotations.TopicDraftNumberOfPollItems;
 
-import javax.validation.constraints.NotNull;
 import javax.validation.constraints.Size;
 
 /**
@@ -32,26 +31,23 @@ import javax.validation.constraints.Size;
 }, message = "{topicDraft.fields.not_null}")
 public class TopicDraft extends Entity {
 
-    @Size(min = Topic.MIN_NAME_SIZE, max = Topic.MAX_NAME_SIZE,
+    @Size(min = 1, max = Topic.MAX_NAME_SIZE,
             message = "{javax.validation.constraints.Size.message}")
     private String title;
 
-    @Size(min = Post.MIN_LENGTH, max = Post.MAX_LENGTH,
+    @Size(min = 1, max = Post.MAX_LENGTH,
             message = "{javax.validation.constraints.Size.message}")
     private String content;
 
-    @NotNull
-    private JCUser topicStarter;
-
-    @NotNull
-    private DateTime lastSaved;
-
-    @Size(min = Poll.MIN_TITLE_LENGTH, max = Poll.MAX_TITLE_LENGTH,
+    @Size(min = 1, max = Poll.MAX_TITLE_LENGTH,
             message = "{javax.validation.constraints.Size.message}")
     private String pollTitle;
 
-    @TopicDraftNumberOfPollItems
+    @TopicDraftNumberOfPollItems(min = 1, max = Poll.MAX_ITEMS_NUMBER)
     private String pollItemsValue;
+
+    private JCUser topicStarter;
+    private DateTime lastSaved;
 
     public TopicDraft() {
     }
--- a/jcommune-model/src/test/java/org/jtalks/jcommune/model/entity/TopicDraftTest.java
+++ b/jcommune-model/src/test/java/org/jtalks/jcommune/model/entity/TopicDraftTest.java
@@ -14,15 +14,12 @@
  */
 package org.jtalks.jcommune.model.entity;
 
-import org.apache.commons.lang.RandomStringUtils;
+import org.apache.commons.lang3.RandomStringUtils;
 import org.joda.time.DateTimeUtils;
 import org.testng.annotations.BeforeMethod;
 import org.testng.annotations.Test;
 
-import javax.validation.ConstraintViolation;
-import javax.validation.Validation;
-import javax.validation.Validator;
-import javax.validation.ValidatorFactory;
+import javax.validation.*;
 import java.util.Set;
 
 import static org.testng.Assert.assertEquals;
@@ -41,7 +38,7 @@ public class TopicDraftTest {
     }
 
     @Test
-    public void validationOfFilledDraftShouldPassSuccessfully() {
+    public void validationOfFilledDraftShouldPass() {
         TopicDraft draft = ObjectsFactory.getDefaultTopicDraft();
 
         Set<ConstraintViolation<TopicDraft>> constraintViolations = validator.validate(draft);
@@ -64,7 +61,7 @@ public class TopicDraftTest {
     }
 
     @Test
-    public void validationOfDraftWithoutTitleShouldPassSuccessfully() {
+    public void validationOfDraftWithoutTitleShouldPass() {
         TopicDraft draft = ObjectsFactory.getDefaultTopicDraft();
 
         draft.setTitle(null);
@@ -75,7 +72,18 @@ public class TopicDraftTest {
     }
 
     @Test
-    public void validationOfDraftWithoutContentShouldPassSuccessfully() {
+    public void validationOfDraftWithTooLongTitleShouldFail() {
+        TopicDraft draft = ObjectsFactory.getDefaultTopicDraft();
+
+        draft.setTitle(RandomStringUtils.random(121));
+
+        Set<ConstraintViolation<TopicDraft>> constraintViolations = validator.validate(draft);
+
+        assertEquals(constraintViolations.size(), 1);
+    }
+
+    @Test
+    public void validationOfDraftWithoutContentShouldPass() {
         TopicDraft draft = ObjectsFactory.getDefaultTopicDraft();
 
         draft.setContent(null);
@@ -86,7 +94,18 @@ public class TopicDraftTest {
     }
 
     @Test
-    public void validationOfDraftWithoutPollTitleShouldPassSuccessfully() {
+    public void validationOfDraftWithTooLongContentShouldFail() {
+        TopicDraft draft = ObjectsFactory.getDefaultTopicDraft();
+
+        draft.setContent(RandomStringUtils.random(20001));
+
+        Set<ConstraintViolation<TopicDraft>> constraintViolations = validator.validate(draft);
+
+        assertEquals(constraintViolations.size(), 1);
+    }
+
+    @Test
+    public void validationOfDraftWithoutPollTitleShouldPass() {
         TopicDraft draft = ObjectsFactory.getDefaultTopicDraft();
 
         draft.setPollTitle(null);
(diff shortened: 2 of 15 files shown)


jcommune-model/src/main/java/org/jtalks/jcommune/model/entity/TopicDraft.java
```diff
@@ -32,27 +31,24 @@
 }, message = "{topicDraft.fields.not_null}")
 public class TopicDraft extends Entity {
 
-    @Size(min = Topic.MIN_NAME_SIZE, max = Topic.MAX_NAME_SIZE,
+    @Size(min = 1, max = Topic.MAX_NAME_SIZE,
```
ctapobep: A thought occurred to me.. Why don't we always have a draft for a person? Even if it's empty. Would that simplify our design?

jcommune-model/src/test/java/org/jtalks/jcommune/model/entity/TopicDraftTest.java
```diff
@@ -119,21 +127,25 @@ public void validationOfDraftWithWrongNumberOfPollItemsShouldFail() {
     }
 
     @Test
-    public void validationOfDraftWithoutTopicStarterShouldFail() {
+    public void validationOfDraftWithoutPollItemsValueShouldPassSuccessFull() {
```
ctapobep: You can remove `SuccessFull` part - the name is clear without it.

jcommune-view/jcommune-web-view/src/main/webapp/WEB-INF/jsp/codeReviewForm.jsp
```diff
@@ -64,4 +64,16 @@
     <spring:message code="label.back"/>
   </a>
 </div>
+<script>
+    $(function() {
+        <c:choose>
+            <c:when test="${topicDraft.id != 0}">
+                window.topicDraft.init($('.keymaps-caption'), <c:out value="${topicDraft.lastSaved.millis}"/>);
```
ctapobep: Can we invoke this method every time and determine if UI should react from within that method? We can determine it if value 0 is passed or undefined.

ctapobep: Also what is 'keymap-caption'? Maybe we could come up with something more reader-friendly..

jcommune-view/jcommune-web-view/src/main/webapp/WEB-INF/jsp/topic/topicForm.jsp
```diff
@@ -143,4 +143,16 @@
     <spring:message code="label.back"/>
   </a>
 </div>
+<script>
+    $(function() {
+        <c:choose>
+            <c:when test="${topicDraft.id != 0}">
+                window.topicDraft.init($('#editorBbCodeDiv'), <c:out value="${topicDraft.lastSaved.millis}"/>);
```
ctapobep: Different locators on different pages, we probably can unify them, can't we?

jcommune-view/jcommune-web-view/src/main/webapp/resources/javascript/app/topicDraft.js
```diff
@@ -0,0 +1,559 @@
...
+        POLL_ITEM_LENGTH_ERROR = $pollItemLength
+            .replace('{min}', MIN_POLL_ITEM_LENGTH).replace('{max}', MAX_POLL_ITEM_LENGTH);
+
+    /**
+     * Class for access to topic draft API
+     *
+     * @constructor
+     */
+    function TopicDraftApi() {
+        this._saveLock = false;
```
ctapobep: Need to describe how these locks are used, what operations they are locking and why we need them. 

Also, isn't having one lock for both save and remove operations makes more sense? If not, please add comments why we need two of them separately.

ctapobep:
Found bugs:
- Draft isn't saved when refreshing or closing the page

jcommune-view/jcommune-web-view/src/main/webapp/resources/javascript/app/topicDraft.js
```diff
@@ -0,0 +1,559 @@
...
+        this._removeLock = false;
+    }
+
+    TopicDraftApi.prototype = {
```
ctapobep: Why overriding `prototype` object? Why not adding functions directly into `TopicDraftApi` or adding functions to the existing `prototype`?

jcommune-view/jcommune-web-view/src/main/webapp/resources/javascript/app/topicDraft.js
```diff
@@ -0,0 +1,559 @@
...
+        /**
+         * Saves or update draft.
+         *
+         * @param data
```
ctapobep: Could you please describe the params?

jcommune-view/jcommune-web-view/src/main/webapp/resources/javascript/app/topicDraft.js
```diff
@@ -0,0 +1,559 @@
...
+    function Timer(fn, intervalTime) {
+        this._fn = fn;
+        this._intervalTime = intervalTime;
+        this._intervalID = null;
+    }
+
+    Timer.prototype = {
+        start: function () {
+            if (this._intervalID === null) {
+                this._intervalID = setInterval(this._fn, this._intervalTime);
```
ctapobep: Mm.. Are you sure we want to use `setInterval()` instead of `setTimeout()`? To me `setTimeout()` should be a default choice since it's more "polite" to the browser and we won't risk with multiple saving processes going right one after each other.

jcommune-model/src/main/java/org/jtalks/jcommune/model/entity/TopicDraft.java
dsdolzhenko: For instance, instead of using `deleteByUser` method of `TopicDraftDao`, we could add method `clean`, which will clean already existing entries. But I don't think that this help us simplify the design.

jcommune-model/src/test/java/org/jtalks/jcommune/model/entity/TopicDraftTest.java
dsdolzhenko: done

jcommune-view/jcommune-web-view/src/main/webapp/resources/javascript/app/topicDraft.js
dsdolzhenko: done

jcommune-view/jcommune-web-view/src/main/webapp/resources/javascript/app/topicDraft.js
dsdolzhenko: In JavaScript there are at least two ways to imitate "normal" classes. First way is define everything in constructor, and second ways is define properties in constructor and methods in prototype. Both of them have own advantages and disadvantages. The second way just is more habitual for me.

jcommune-view/jcommune-web-view/src/main/webapp/resources/javascript/app/topicDraft.js
dsdolzhenko: I thought once again and removed these variables. I don't think that we need them here.

jcommune-view/jcommune-web-view/src/main/webapp/resources/javascript/app/topicDraft.js
dsdolzhenko: This way is also legal. Original prototype of TopicDraftApi is instance of Object, and here I just have created object with additional methods.

jcommune-view/jcommune-web-view/src/main/webapp/resources/css/app/application.css
```diff
@@ -598,6 +598,12 @@ div.signup {
     font-style: italic;
 }
 
+.topic-draft-counter {
```
ctapobep: I dunno what exactly is the reason, but with new changes the space between topic subject and Sticked checkbox increased.

jcommune-view/jcommune-web-view/src/main/webapp/resources/javascript/app/topicDraft.js
```diff
@@ -0,0 +1,534 @@
...
+            .replace('{min}', MIN_POLL_ITEM_LENGTH).replace('{max}', MAX_POLL_ITEM_LENGTH);
+
+    /**
+     * Class for access to topic draft API
+     *
+     * @constructor
+     */
+    function TopicDraftApi() {}
+
+    TopicDraftApi.prototype = {
```
ctapobep: You're still using inheritance incorrectly =)) Should be:
`TopicDraftApi.prototype.save = function() {}`

jcommune-view/jcommune-web-view/src/main/webapp/resources/javascript/app/topicDraft.js
```diff
@@ -0,0 +1,534 @@
...
+     * Helper for executing functions at specified interval
+     *
+     * @param {Function} fn the called function
+     * @param {Number} delay the interval in milliseconds
+     * @constructor
+     */
+    function IntervalTimer(fn, delay) {
+        this._fn = fn;
+        this._delay = delay;
+        this._intervalID = null;
```
ctapobep: Add docs to these fields please.

jcommune-view/jcommune-web-view/src/main/webapp/resources/javascript/app/topicDraft.js
```diff
@@ -0,0 +1,534 @@
...
+    }
+
+    IntervalTimer.prototype = {
+        start: function () {
+            if (this._intervalID === null) {
+                this._intervalID = setInterval(this._fn, this._delay);
```
ctapobep: Still using setInterval() instead of setTimeout().

jcommune-view/jcommune-web-view/src/main/webapp/resources/javascript/app/topicDraft.js
```diff
@@ -0,0 +1,534 @@
...
+     * @constructor
+     */
+    function Counter() {
+        this._el = $("<span id='topicDraftCounter' class='topic-draft-counter pull-right'></span>");
+        this._interval = 5 * SECOND;
+        this._timer = new IntervalTimer(this._tick.bind(this), this._interval);
+        this._time = 0;
+    }
+
+    Counter.prototype = {
```
ctapobep: Same problem with the prototype.

jcommune-view/jcommune-web-view/src/main/webapp/resources/javascript/app/topicDraft.js
```diff
@@ -0,0 +1,534 @@
...
+        restart: function () {
+            this.stop();
+            this.start();
+        },
+
+        getElement: function () {
+            return this._el;
+        },
+
+        _tick: function () {
```
ctapobep: Docs required.

jcommune-view/jcommune-web-view/src/main/webapp/resources/javascript/app/topicDraft.js
```diff
@@ -0,0 +1,534 @@
...
+                error: (function() {
+                    var error = $('#topic\\.poll\\.pollItems');
+
+                    return error.length ? error
+                                        : defaultError.clone().insertAfter(self._pollItemsValue)
+                })()
+            }
+        };
+
+        this._wasChanged = false;
```
ctapobep: Could you please add docs here? Also, we need to put all the fields in one place (at the top or at the bottom).

jcommune-view/jcommune-web-view/src/main/webapp/resources/javascript/app/topicDraft.js
```diff
@@ -0,0 +1,534 @@
...
+                this.save().then(function () {
+                    self._wasChanged = false;
+                });
+            }
+        },
+
+        _onKeyUp: function (event) {
+            var self = this;
+
+            this._wasChanged = true;
```
ctapobep: This means that the rules about when the draft is getting saved are changed. Previously we were checking if the actual data were changed and only if it did we were sending the request to the server. I'm more or less okay with your approach, but we should notify the team about this decision in the comments to the JIRA issue.

jcommune-view/jcommune-web-view/src/main/webapp/resources/css/app/application.css
dsdolzhenko: The problem was not in this css rule. But now it's fixed.

ctapobep:
@dsdolzhenko could you squash the commits into 1 and I'll merge them?

dsdolzhenko:
done

PR #143, opened 2015-10-21
JC-2214 Get rid of Back button on Post/Topic Editing page

SurfVaporizer:
- back button has been removed from Post editing page
- back button has been removed from Topic editing page


ctapobep:
@SurfVaporizer what about i18n for that button?

SurfVaporizer:
@ctapobep Should I delete (label.back) field? Or what do you mean?
btw, field label.back=Back used in another parts of that project

ctapobep:
Ok, if so there there is no need to delete that label.

ctapobep:
Merged

PR #144, opened 2015-10-25
JC-2018 text of "Recent activity" and "Search results" have two rows

vps2:
- replaced classes responsible for a dimension elements "div".


vps2:
Please tell me: when you fix bugs in UI, can I use HTML5 tags (e.g., autofocus) or should I stick with HTML4?

ctapobep:
@dsdolzhenko what do you think?

dsdolzhenko:
@ctapobep @vps2, I think, there is no reason not to use HTML5 features today. Moreover, we already use them in this project.

vps2:
Understood. 
Thanks for the reply.

PR #145, opened 2015-10-26

dsdolzhenko:
- refactor postDraft using parts from topicDraft

### DIFF
--- a/jcommune-service/src/main/java/org/jtalks/jcommune/service/transactional/TransactionalTopicDraftService.java
+++ b/jcommune-service/src/main/java/org/jtalks/jcommune/service/transactional/TransactionalTopicDraftService.java
@@ -41,6 +41,7 @@ public class TransactionalTopicDraftService implements TopicDraftService, Plugin
      * {@inheritDoc}
      */
     @Override
+    @PreAuthorize("isAuthenticated()")
     public TopicDraft getDraft() {
         JCUser user = userService.getCurrentUser();
         return topicDraftDao.getForUser(user);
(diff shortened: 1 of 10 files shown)


jcommune-service/src/main/java/org/jtalks/jcommune/service/transactional/TransactionalTopicDraftService.java
```diff
@@ -61,8 +62,10 @@ public TopicDraft saveOrUpdateDraft(TopicDraft draft) {
         } else {
             currentDraft.setContent(draft.getContent());
             currentDraft.setTitle(draft.getTitle());
-            currentDraft.setPollTitle(draft.getPollTitle());
-            currentDraft.setPollItemsValue(draft.getPollItemsValue());
+            if (draft.getPollTitle() != null || draft.getPollItemsValue() != null) {
```
ctapobep: Why do we need this condition? Isn't contentDraft.pollTitle null anyway?

jcommune-view/jcommune-web-view/src/main/webapp/resources/javascript/app/postDraft.js
```diff
@@ -13,405 +13,255 @@
...
-        }
-        startTimer();
-        if (getTextLength() == 0 && draftState.previouslySavedText.length !== 0) {
-          draftState.previouslySavedText = '';
-          deleteDraft();
+    PostDraft.prototype.save = function (async) {
+        var self = this,
+            draft = this._getDraftState();
+
+        async = async !== false;
```
ctapobep: Nice trick :)

jcommune-view/jcommune-web-view/src/main/webapp/resources/javascript/app/postDraft.js
```diff
@@ -13,405 +13,255 @@
...
+    /**
+     * Determines whether the post was changed from last saving.
+     *
+     * @returns {boolean} whether this draft was changed
+     */
+    PostDraft.prototype.wasChanged = function () {
+        var currentState = this._getDraftState(),
+            previousState = this._lastSavedDraftState;
+
+        return currentState['bodyText'] !== previousState['bodyText'];
```
ctapobep: Because the body can be pretty large, can we first check the length and only if matches compare the contents? I assume that should be much faster.

jcommune-service/src/main/java/org/jtalks/jcommune/service/transactional/TransactionalTopicDraftService.java
dsdolzhenko: We need this condition because when we save draft that does not contain poll, we should not overwrite already exists values. And now, when user removes the text from field, instead of sending null, we will send empty string.

I have added comment to this condition in the next commit.

jcommune-view/jcommune-web-view/src/main/webapp/resources/javascript/app/postDraft.js
dsdolzhenko: I think that javascript engine already do this under hood when comparing strings.

ctapobep:
Could you please squash your commits into 1?

dsdolzhenko:
done

PR #147, opened 2015-10-28
JC-2268 Focus on sign in/sign up pages should be on first input field

vps2:
- added a code that selects the input field with the attribute
  "autofocus" of the form when page is loaded. Made to support
  the set focus on the element with attribute "autofocus" in browsers
  that do not support HTML5.
- added the attribute "autofocus" to the field "user name" on the pages:
  registration, login.
- added the attribute "autofocus" to the field "user email" on the password
  recovery page.


ctapobep:
@SurfVaporizer could you please review the changes? Review includes both code review and testing the change locally.

SurfVaporizer:
Looks good, tested with Chrome and FF on Windwos platform.

PR #148, opened 2015-10-29
JC-2102 Add unique key (topic_id, user_id) to LAST_READ_POSTS table

dsdolzhenko:
see comments for [JC-2102](http://jira.jtalks.org/browse/JC-2102?focusedCommentId=30100&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-30100)


ctapobep:
As far as I remember if we have a compound key, it's important that we search in the same order the key declares columns. Could you please double check that and ensure that in our search queries the order is correct? I'd also leave a comment at those queries that the order is important due to...

dsdolzhenko:
Yes, you right. MySQL [does not guarantee](https://dev.mysql.com/doc/refman/5.7/en/multiple-column-indexes.html) usage of index to perform lookups if the columns do not form a leftmost prefix of the index.

Currently, all queries to this table use the right order of columns.

ctapobep:
Ok, thanks for the research and the updates. I'll merge this change when we're done with performance testing.

ctapobep:
Merged

PR #151, opened 2015-11-03
JC-2051 password changed on registering user from sign up page

vps2:
- added code that returns the original (not encoded) password into UserDto
  object when the registration failed.

### DIFF
--- a/jcommune-service/src/main/java/org/jtalks/jcommune/service/transactional/TransactionalAuthenticator.java
+++ b/jcommune-service/src/main/java/org/jtalks/jcommune/service/transactional/TransactionalAuthenticator.java
@@ -324,8 +324,9 @@ public class TransactionalAuthenticator extends AbstractTransactionalEntityServi
         BindingResult jcErrors = new BeanPropertyBindingResult(registerUserDto, "newUser");
         validator.validate(registerUserDto, jcErrors);
         UserDto userDto = registerUserDto.getUserDto();
-        String encodedPassword = (userDto.getPassword() == null || userDto.getPassword().isEmpty()) ? ""
-                : encryptionService.encryptPassword(userDto.getPassword());
+        String notEncodedPassword = userDto.getPassword();
+        String encodedPassword = (notEncodedPassword == null || notEncodedPassword.isEmpty()) ? ""
+                : encryptionService.encryptPassword(notEncodedPassword);
         userDto.setPassword(encodedPassword);
         registerByPlugin(userDto, true, result);
         mergeValidationErrors(jcErrors, result);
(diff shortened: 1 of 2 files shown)


ctapobep:
Hi, @vps2 would you please also add unit or component tests for this?
@dsdolzhenko could you please review/test the changes?

vps2:
I have a question: recover the original password should only occur when there is an BindingResult.hasError() return true in the registration method or the same thing must be done, if the exception occurred?
Tests implementation depends from this.

ctapobep:
If exception happens, we show Error 500, so the form is not visible anyway.

vps2:
- added tests.

P/S: in my opinion, in tests of TransactionalAuthenticatorTest class, which includes code like this: "when(bindingResult.hasErrors())" - it doesn't matter, because the method is never called. And some tests are not valid because the validation condition "verify(bindingResult, never()).rejectValue(anyString(), anyString(), anyString());" is always true.

vps2:
up

ctapobep:
@vps2 if there are tests that you think are written badly, feel free to remove/update them. We're doing this from time to time.
As per your new tests:
- They are pretty complicated and long - usually tests shouldn't take more than 5 lines
- There is a lot of mocking - do you think we could find ways of minimizing the mocking? The idea behind having as few mocks as possible is described in these places: [http://qala.io/blog/test-pyramid.html](http://qala.io/blog/test-pyramid.html), [Ian Cooper: TDD, where did it all go wrong](https://vimeo.com/68375232) and [Jimmy Bogard: Holistic testing](https://vimeo.com/68390508). Also you could read/participate in [this discussion](http://javatalks.ru/topics/47966). We also started to write Component Tests e.g. [here](https://github.com/jtalks-org/jcommune/blob/f455a3f189340c39b89450c4d8633a0551bb1705/jcommune-view/jcommune-web-view/src/test/java/org/jtalks/jcommune/test/TopicControllerTest.groovy#L85-L85).

If you feel overwhelmed, give me a shout, I'll try to help you :)

vps2:
I think that help can not hurt.
Need something to do with validator. Also, i think that from the mocks of "pluginService" and "registrationPlugin" is not to get rid of.

ctapobep:
Ok, let's leave it as it is right now. Your new changes are a good improvement over the initial version. Could you please squash the changes into 1 commit?

We still need to gradually move towards improving our testing, so keep reading the aforementioned literature and ask questions if you have any. You may use javatalks.ru (preferably) or you may ask questions in PRs.

vps2:
Did.

ctapobep:
Merged

PR #154, opened 2015-11-12

dsdolzhenko:
bug fixes for:

* [Error message about size in Topic title field isn`t disappeared after editing](http://jira.jtalks.org/browse/JC-2201?focusedCommentId=30171&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-30171)


ctapobep:
Could you do some corrections on the commit message:
- First line should be more distinguishable. E.g. `JC-2201 Fix validation error for title/poll items doesn't disappear`
- The rest should also describe why the error happened and/or how the fix helped to fix it.

PR #155, opened 2015-11-21
JC-2134 The user's selection does not work by Enter in Firefox

vps2:
- fixed suggestion behavior in firefox.

P/S 1: found duplicated issue - JC-2213.
P/S 2: in my opinion, the below code in signin.js is redundant because the element '#restorePassSuccess' is missing (or is it groundwork for the future?):

    var success = $('#restorePassSuccess');
    if (success.length > 0) {
        var bodyContent = '\
            <div id="restore-passwd" class="control-group"> \
                <h4>' + success.val() + '</h4> \
            </div>';

        var footerContent = '<button id="restore-ok-button" class="btn btn-primary" name="confirm"> \
            ' + $labelOk + '</button>';

        var goToLoginPage = function (e) {
            if (e) {
                e.preventDefault();
            }
            window.location.href = $('.brand').attr('href') + 'login';
        };

        jDialog.createDialog({
            dialogId: 'restore-password-modal-dialog',
            closeDialog: goToLoginPage,
            bodyContent: bodyContent,
            footerContent: footerContent,
            maxWidth: 350,
            tabNavigation: ['#restore-ok-button'],
            handlers: {
                '#restore-ok-button': {'click': goToLoginPage}
            }
        });
    }

### DIFF
--- a/jcommune-view/jcommune-web-view/src/main/webapp/resources/javascript/app/dialog.js
+++ b/jcommune-view/jcommune-web-view/src/main/webapp/resources/javascript/app/dialog.js
@@ -322,7 +322,8 @@ $(function () {
         var tabNavigation = function (selectors) {
             $.each(selectors, function (idx, v) {
                 var func = function (e) {
-                    if ((e.keyCode || e.charCode) === tabCode) {
+                    //The "jQuery.which" property does not work on the "onkeypress" event in Firefox.
+                    if ((e.which || e.keyCode) === tabCode) {
                         e.preventDefault();
                         // to prevent infinite cycle if all element are not visible or do not exist
                         var skipCount = 0;
@@ -336,12 +337,22 @@ $(function () {
                             }
                             nextElement = jDialog.dialog.find(selectors[currentIndex]);
                             ++skipCount;
-                        } while ((nextElement.length === 0 || !nextElement.is(":visible"))
+                        } while ((nextElement.length === 0 || nextElement.is(":not(:visible)"))
                                  && skipCount <= selectors.length);
                         nextElement.focus();
                     }
                 };
-                $(v).on('keydown', func);
+
+                if($.browser.mozilla) {
+                    /*
+                    In Firefox overriding the "keydown" method with "event.preventDefault()" code inside, stops
+                    working autocomplete when pressing Tab (and another default behavior).
+                    Instead of overriding "keydown", you can use the "keypress" method.
+                     */
+                    $(v).on('keypress', func);
+                } else {
+                    $(v).on('keydown', func);
+                }
             });
         };
 
--- a/jcommune-view/jcommune-web-view/src/main/webapp/resources/javascript/app/signin.js
+++ b/jcommune-view/jcommune-web-view/src/main/webapp/resources/javascript/app/signin.js
@@ -40,21 +40,23 @@ $(function () {
         var footerContent = '<input  type="submit" id="signin-submit-button" value="' + $labelSignin + '" class="btn btn-primary" name="commit"/>';
 
         var submitDialog = function (e) {
-            if (e.keyCode == enterCode) {
-                //if focus on username then select password field
-                if ($(e.target).is('#userName')) {
+            if(e.which == enterCode) {
+                /*
+                 Simulate submit when enter pressed on "input" element in the internet explorer.
+
+                 Note: in internet explorer (specifically ie8) when you press "Enter" button on the item "input", does
+                 not cause an event "submit" of the form. If we move focus to the submit button and press "Enter" (when
+                 form fields filled with error), then press "Enter" in the next time on the input element ("username")
+                 will cause this event. But there is one comment: although visually focused will remain "username"
+                 field, but actually - form. Therefore, in the "isPreventSubmitFor($inputElement)" method will be passed
+                 an incorrect value. Consequently, the method will produce incorrect result.
+                 */
+                if($.browser.msie && $(e.target).is(':input')) {
                     e.preventDefault();
-                    if ($.browser.mozilla) {
-                        setTimeout(function () {
-                            jDialog.dialog.find('#password').focus();
-                        }, 0);
-                    }
-                    else {
-                        jDialog.dialog.find('#password').focus();
-                    }
+                    sendLoginPost(e);
                 }
             }
-            if ((e.keyCode || e.charCode) == escCode) {
+            if (e.which == escCode) {
                 jDialog.dialog.find('.close').click();
             }
         };
@@ -75,7 +77,8 @@ $(function () {
                 }}
 
             },
-            dialogKeydown: submitDialog
+            dialogKeydown: submitDialog,
+            preventSubmitInputElements: ['#userName']
         });
     });
 
@@ -111,29 +114,29 @@ $(function () {
 });
 
 function sendEmailConfirmation(recipient) {
-  $.ajax({
-          type: 'GET',
-          url: $root + '/confirm?id='+recipient,
-          success: function () {
-          var message = "";
-             if (resp.status == 'SUCCESS') {
+    $.ajax({
+        type: 'GET',
+        url: $root + '/confirm?id='+recipient,
+        success: function () {
+            var message = "";
+            if (resp.status == 'SUCCESS') {
                 message = $labelEmailConfirmationWasSent;
-              } else {
+            } else {
                 message = $labelError500Detail
-              }
-                jDialog.createDialog({
-                    type: jDialog.alertType,
-                    bodyMessage: $labelEmailConfirmationWasSent
-              });
-          },
-          error: function (jqXHR, textStatus, errorThrown) {
+            }
             jDialog.createDialog({
-               type: jDialog.alertType,
-               bodyMessage: $labelError500Detail
-             });
-          }
-  });
-};
+                type: jDialog.alertType,
+                bodyMessage: $labelEmailConfirmationWasSent
+            });
+        },
+        error: function (jqXHR, textStatus, errorThrown) {
+            jDialog.createDialog({
+                type: jDialog.alertType,
+                bodyMessage: $labelError500Detail
+            });
+        }
+    });
+}
 
 /**
  * Handles submit request from login form by sending POST request, with params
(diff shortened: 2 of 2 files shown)


jcommune-view/jcommune-web-view/src/main/webapp/resources/javascript/app/signin.js
```diff
@@ -40,18 +40,11 @@ $(function () {
         var footerContent = '<input  type="submit" id="signin-submit-button" value="' + $labelSignin + '" class="btn btn-primary" name="commit"/>';
 
         var submitDialog = function (e) {
-            if (e.keyCode == enterCode) {
-                //if focus on username then select password field
-                if ($(e.target).is('#userName')) {
+            if((e.keyCode || e.charCode) == enterCode) {
+                //simulate submit on input[type='text'] enter in ie
+                if($.browser.msie && $(e.target).is(':input')) {
```
ctapobep: Could you please explain in comments what this is and why it's needed?

jcommune-view/jcommune-web-view/src/main/webapp/resources/javascript/app/signin.js
```diff
@@ -40,18 +40,11 @@ $(function () {
         var footerContent = '<input  type="submit" id="signin-submit-button" value="' + $labelSignin + '" class="btn btn-primary" name="commit"/>';
 
         var submitDialog = function (e) {
-            if (e.keyCode == enterCode) {
-                //if focus on username then select password field
-                if ($(e.target).is('#userName')) {
+            if((e.keyCode || e.charCode) == enterCode) {
```
ctapobep: According to this http://www.w3schools.com/jsref/event_key_charcode.asp it's e.key that should be used instead of `e.keyCode` and `e.charCode`. But the latter should still be checked for backward compatibility. 

Could you please create a global function that can return the key code and it checks all the 3 options? And also leave comments that some of the options are used for backward compatibility.

jcommune-view/jcommune-web-view/src/main/webapp/resources/javascript/app/signin.js
```diff
@@ -143,7 +137,15 @@ function sendEmailConfirmation(recipient) {
  * get error message, providing user with opportunity to change login or password
  */
 function sendLoginPost(e) {
+    var $currentFocusedElement = $('input:focus');
+    if(isPreventSubmitFor($currentFocusedElement))
+    {
```
ctapobep: This `{` should be kept on the previous line.

ctapobep:
@vps2 could you please add into the commit message what was wrong and how you fixed it?

ctapobep:
@vps2 if you feel that some chunk can be removed, you could send deletion as a separate commit with explanations.

jcommune-view/jcommune-web-view/src/main/webapp/resources/javascript/app/signin.js
vps2: Remember that jquery has the desired property "event.which". 
Instead custom global function will be better to use it's.

ctapobep: That's much better, agree :)

vps2:
Note:
- in the method "submitDialog()" invocation the "e.preventDefault()" didn't give normally to work the suggestion list in input box in browser Firefox. If invocation to move lower in the block "else", then suggestion list starts to work correctly, but there is another problem - with a simple (without a suggestion list) pressing the "Enter" button on the field "#userName" calls the event "submit" and not advance to the next field. So I decided to handle the transition to the next field in a method that is invoked in response to an event "submit".
- in internet explorer (specifically ie8) when you press "Enter" button on the item "input", event of "submit" doesn't work constantly, and this element loses focus. So in the "submitDialog" i call the appropriate method manually.
- variable - array "preventSubmitElements" is used to define the "input" elements of the form, while pressing "Enter" on which a form should not be sent to the server. If "Enter" is pressed on such item, then instead of sending the form to the server the focus from it will be moved to the next item.

ctapobep:
@vps2 I'll look into this in more details a bit later, but what I noticed right now is that Tab (which should select the value from the list _and_ move to the next field) doesn't work in FF as well - it doesn't choose the selected value from the dropdown. 

Can we fix it in this PR too?

vps2:
If i understand correctly, this is the default behavior in the firefox:
https://support.mozilla.org/ru/kb/upravlenie-avtomaticheskim-zapolneniem-form-vashej

vps2:
>but what I noticed right now is that Tab (which should select the value from the list and move to the next field) doesn't work in FF as well - it doesn't choose the selected value from the dropdown.
>Can we fix it in this PR too?

Also, as in the problem with the "Enter" here is to blame invokation of "e.preventDefault() in the method "tabNavigation" (file: dialog.js).
But if it is removed, then the tabulation break.

vps2:
Fixed incorrect behaviour of autocomplete with the Tab key in Firefox.

ctapobep:
@vps2 Thanks! I'll take a look in the nearest future.

ctapobep:
@res5515 could you please review and test these changes?

jcommune-view/jcommune-web-view/src/main/webapp/resources/javascript/app/dialog.js
```diff
@@ -322,7 +322,8 @@ $(function () {
         var tabNavigation = function (selectors) {
             $.each(selectors, function (idx, v) {
                 var func = function (e) {
-                    if ((e.keyCode || e.charCode) === tabCode) {
+                    //The "jQuery.which" property does not work on the "onkeypress" event in Firefox.
+                    if ((e.which || e.keyCode) === tabCode) {
                         e.preventDefault();
```
ctapobep: I assume this comment from the commit message should be placed here:
>in internet explorer (specifically ie8) when you press "Enter" button on the item "input", event of "submit" doesn't work constantly, and this element loses focus. So in the "submitDialog" i call the appropriate method manually.

But I don't understand what "doesn't work constantly" means, could you please rephrase a bit and place an inline comment here.

jcommune-view/jcommune-web-view/src/main/webapp/resources/javascript/app/signin.js
```diff
@@ -211,4 +212,28 @@ function sendLoginPost(e) {
             });
         }
     });
-};
\ No newline at end of file
+}
+
+function isPreventSubmitFor($element) {
```
ctapobep: Could you please add docs here about the purpose of this method?

jcommune-view/jcommune-web-view/src/main/webapp/resources/javascript/app/signin.js
```diff
@@ -211,4 +212,28 @@ function sendLoginPost(e) {
...
+            if ($element.is(value)) {
+                isPreventFormSubmit = true;
+                return false;
+            }
+        });
+
+    return isPreventFormSubmit;
+}
+
+function navigateToNext($currentElement) {
```
ctapobep: And here as well.

jcommune-view/jcommune-web-view/src/main/webapp/resources/javascript/app/signin.js
```diff
@@ -211,4 +212,28 @@ function sendLoginPost(e) {
...
         }
     });
-};
\ No newline at end of file
+}
+
+function isPreventSubmitFor($element) {
+    var isPreventFormSubmit = false;
+
+    $.each(jDialog.options.preventSubmitElements, function (idx, value) {
```
ctapobep: Could we unwrap it into usual JS iteration (with `for(){}`)? It will be 1 additional line, but we'll unbind ourselves from JQuery (on the off chance we'll be migrating from it or from its current version). 

There is also a similar native function: [Array.prototype.forEach](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach) which does a similar thing. Won't work in IE8 though.

jcommune-view/jcommune-web-view/src/main/webapp/resources/javascript/app/dialog.js
```diff
@@ -336,12 +337,19 @@ $(function () {
...
                         nextElement.focus();
                     }
                 };
-                $(v).on('keydown', func);
+
+                if($.browser.mozilla) {
+                    //To correct the problem with incorrect behaviour of autocomplete with the Tab key in Firefox.
+                    $(v).on('keypress', func);
+                }
+                else {
```
ctapobep: Formatting: should be placed on the previous line.

jcommune-view/jcommune-web-view/src/main/webapp/resources/javascript/app/dialog.js
```diff
@@ -336,12 +337,19 @@ $(function () {
...
-                        } while ((nextElement.length === 0 || !nextElement.is(":visible"))
+                        } while ((nextElement.length === 0 || nextElement.is(":not(:visible)"))
                                  && skipCount <= selectors.length);
                         nextElement.focus();
                     }
                 };
-                $(v).on('keydown', func);
+
+                if($.browser.mozilla) {
+                    //To correct the problem with incorrect behaviour of autocomplete with the Tab key in Firefox.
```
ctapobep: Could you please add more details for the future generation on why it's not working and how the fix helps?

jcommune-view/jcommune-web-view/src/main/webapp/resources/javascript/app/signin.js
```diff
@@ -40,21 +40,14 @@ $(function () {
         var footerContent = '<input  type="submit" id="signin-submit-button" value="' + $labelSignin + '" class="btn btn-primary" name="commit"/>';
 
         var submitDialog = function (e) {
-            if (e.keyCode == enterCode) {
-                //if focus on username then select password field
-                if ($(e.target).is('#userName')) {
+            if(e.which == enterCode) {
+                //simulate submit when enter pressed on "input" element in the internet explorer
+                if($.browser.msie && $(e.target).is(':input')) {
                     e.preventDefault();
```
ctapobep: I assume this comment from the commit message should be placed here:
>in internet explorer (specifically ie8) when you press "Enter" button on the item "input", event of "submit" doesn't work constantly, and this element loses focus. So in the "submitDialog" i call the appropriate method manually.

But I don't understand what "doesn't work constantly" means, could you please rephrase a bit and place an inline comment here.

ctapobep:
Tested on Chrome and FF on Mac - seems to be working fine.

vps2:
Corrected previous remarks.

ctapobep:
Looks perfect, could you please squash the commits?

vps2:
Did.

ctapobep:
Unfortunately while you were working on your changes someone else has changed those files as well so there are conflicts. Would you be able to rebase your commit on our master? Sorry for being so slow.

vps2:
Done.

PR #158, opened 2015-11-26
When "isJSCompressed" is false, then webpages is displayed wrong

vps2:
When in the root pom file property "isJsCompressed" is set to "false", the web application must not use compressed resources. As a result of the use of such resources - is that webpages is not displayed correctly.
This change fixes this problem.

### DIFF
--- a/jcommune-view/jcommune-web-view/pom.xml
+++ b/jcommune-view/jcommune-web-view/pom.xml
@@ -192,6 +192,50 @@
           </deployer>
         </configuration>
       </plugin>
+      <!--This plugin used to specify resources, that need include into war file depending
+      on the value of property "isJsCompressed"-->
+      <plugin>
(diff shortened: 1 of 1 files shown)


ctapobep:
My thoughts on the subject:
- Locally when building directly from IntelliJ (as opposed to Maven) it won't run maven plugins to compress them. So locally the compression shouldn't be a problem.
- We don't quite need uncompressed JS files even for non-local envs - instead we can generate *.map files (uncompressed version of the same files) that Browser Dev Tools would find and open instead of compressed files. I just created [JC-2273](http://jira.jtalks.org/browse/JC-2273) for this purpose.
- Even if we need to pick different values of the same property depending on other properties - in Maven it's easier to do with profiles than with Ant Plugin.

So my suggestion is to abandon this PR and rather take JC-2273 (either now or at some point in the future).

vps2:
Ok.

PR #164, opened 2015-12-21
JC-2276 Disable email notification after deleting post

vps2:
- when you delete a post, a notification is sent only to the user who
  created it. If the user (owner of the post) deletes the post, no notification is sent.

### DIFF
--- a/jcommune-service/src/main/java/org/jtalks/jcommune/service/nontransactional/NotificationService.java
+++ b/jcommune-service/src/main/java/org/jtalks/jcommune/service/nontransactional/NotificationService.java
@@ -85,17 +85,16 @@ public class NotificationService {
     /**
      * Overload for skipping topic subscribers
      *
-     * @param entity
-     * @param topicSubscribers
+     * @param entity changed subscribed entity.
+     * @param excludeFromNotification a collection of subscribers who are excluded from a notification.
      */
-    public void subscribedEntityChanged(SubscriptionAwareEntity entity, Collection<JCUser> topicSubscribers) {
+    public void subscribedEntityChanged(SubscriptionAwareEntity entity, Collection<JCUser> excludeFromNotification) {
         Collection<JCUser> subscribers = subscriptionService.getAllowedSubscribers(entity);
+        subscribers.removeAll(excludeFromNotification);
         filterSubscribers(subscribers, entity);
 
         for (JCUser user : subscribers) {
-            if (!topicSubscribers.contains(user)) {
-                mailService.sendUpdatesOnSubscription(user, entity);
-            }
+            mailService.sendUpdatesOnSubscription(user, entity);
         }
     }
 
--- a/jcommune-service/src/main/java/org/jtalks/jcommune/service/transactional/TransactionalPostService.java
+++ b/jcommune-service/src/main/java/org/jtalks/jcommune/service/transactional/TransactionalPostService.java
@@ -46,7 +46,8 @@ import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
-
+import java.util.Set;
+import java.util.HashSet;
 
 /**
  * Post service class. This class contains method needed to manipulate with Post persistent entity.
@@ -139,8 +140,8 @@ public class TransactionalPostService extends AbstractTransactionalEntityService
             "(hasPermission(#post.topic.branch.id, 'BRANCH', 'BranchPermission.DELETE_OTHERS_POSTS') and " +
             "#post.userCreated.username != principal.username)")
     public void deletePost(Post post) {
-        JCUser user = post.getUserCreated();
-        user.setPostCount(user.getPostCount() - 1);
+        JCUser postCreator = post.getUserCreated();
+        postCreator.setPostCount(postCreator.getPostCount() - 1);
         Topic topic = post.getTopic();
         topic.removePost(post);
         Branch branch = topic.getBranch();
@@ -149,11 +150,18 @@ public class TransactionalPostService extends AbstractTransactionalEntityService
             branch.clearLastPost();
         }
 
-
         // todo: event API?
         topicDao.saveOrUpdate(topic);
         securityService.deleteFromAcl(post);
-        notificationService.subscribedEntityChanged(topic);
+
+        /*
+         implementation of the task JC-2276 (only the creator of the post should
+         be notified when it's removed).
+         */
+        Set<JCUser> excludeFromNotification = new HashSet<>(topic.getSubscribers());
+        excludeFromNotification.remove(postCreator);
+        notificationService.subscribedEntityChanged(topic, excludeFromNotification);
+
         if (deletedPostIsLastPostInBranch) {
             branchLastPostService.refreshLastPostInBranch(branch);
         }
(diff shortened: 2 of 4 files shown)


jcommune-service/src/main/java/org/jtalks/jcommune/service/transactional/TransactionalPostService.java
```diff
@@ -149,11 +150,14 @@ public void deletePost(Post post) {
...
         }
 
-
         // todo: event API?
         topicDao.saveOrUpdate(topic);
         securityService.deleteFromAcl(post);
-        notificationService.subscribedEntityChanged(topic);
+
+        Set<JCUser> topicSubscribers = new HashSet<>(topic.getSubscribers());
+        topicSubscribers.remove(user);
```
ctapobep: @vps2 the intention of the task is so that only the user who left that post needs to be notified about its removal. Your changes do the opposite - it'll notify everyone except the post creator :)

jcommune-service/src/main/java/org/jtalks/jcommune/service/transactional/TransactionalPostService.java
```diff
@@ -149,11 +150,14 @@ public void deletePost(Post post) {
...
+        notificationService.subscribedEntityChanged(topic, topicSubscribers);
```
vps2: This method overloaded. It notifies only those users that are not in the "topicSubscribers" list.
I.e. with the changes being made will be notified only the creator of the post, if he did not remove himself.

ctapobep: Could you then also rename the parameter of subscribedEntityChanged() and the variable itself from `topicSubscribers` to `excludeFromNotification`? It's misleading in this case.

ctapobep: Additionally I'd leave a comment or javadoc mentioning why we send the notification only to post creator. You could mention the task ID in that comment.

NCNecros: Maybe  we need to add a overloaded method which accept only a user? Foreach for subscribers is a bit slowly than work with one user.

vps2:
I think that from the access to the database (subscribtion service) not to get rid of, because the creator of the post can be unsubscribed from notification. So send him a notice about the removal of his post, it makes no sense.  Or are you suggesting to use the "topic.getSubscriberts()" method instead of "subscriptionService.getAllowedSubscribers()"?

jcommune-service/src/main/java/org/jtalks/jcommune/service/transactional/TransactionalPostService.java
vps2: I think that from the access to the database (subscribtion service) not to get rid of, because the creator of the post can be unsubscribed from notification. So send him a notice about the removal of his post, it makes no sense. Or are you suggesting to use the "topic.getSubscriberts()" method instead of "subscriptionService.getAllowedSubscribers()"?

ctapobep: Agree with @vps2 - the user may be unsubscribed. We may introduce a method `subscriptionService.entityChanged(entity, userToNotify)`, I'm just thinking if this will make things more unclear since with that we'll be finding subscribers in different places on different layers.

NCNecros: OK, I understood.

vps2:
- renamed a local variable and a method parameter, and added a comment
  with task ID.

jcommune-service/src/main/java/org/jtalks/jcommune/service/nontransactional/NotificationService.java
```diff
@@ -85,15 +85,15 @@ public void subscribedEntityChanged(SubscriptionAwareEntity entity) {
...
-     * @param topicSubscribers
+     * @param entity changed subscribed entity.
+     * @param excludeFromNotification a collection of subscribers who are excluded from a notification.
      */
-    public void subscribedEntityChanged(SubscriptionAwareEntity entity, Collection<JCUser> topicSubscribers) {
+    public void subscribedEntityChanged(SubscriptionAwareEntity entity, Collection<JCUser> excludeFromNotification) {
         Collection<JCUser> subscribers = subscriptionService.getAllowedSubscribers(entity);
         filterSubscribers(subscribers, entity);
 
         for (JCUser user : subscribers) {
```
ctapobep: Just figured.. Why don't we do `subscribers.removeAll(excludeFromNotification)`?

ctapobep:
@vps2 could you add one improvement that I mentioned in the last comment and squash the changes into 1 commit?

vps2:
I was wondering about that too. Also need to fix the test.

vps2:
Did one commit.

ctapobep:
Merged, thanks!

vps2:
Will raise the issue above, which have not received a response:
After commit, solving the problem of "JC-2278 Add caching to avatars" i have does not pass the test "AvatarControllerTest.renderAvatar_mustReturnNotModifiedStatus_ifAvatarWasNotChangedSinceLastTime()". How to be?

ctapobep:
That was fixed already, have you pulled the latest master?

vps2:
Yes.
Just now, when building the project, received a broken test.

ctapobep:
Hm.. Weird. What OS do you have and what error do you get?

vps2:
VMWare virtual machine with Windows 7 SP1 x32.

[TEST-org.jtalks.jcommune.web.controller.AvatarControllerTest.zip](https://github.com/jtalks-org/jcommune/files/74824/TEST-org.jtalks.jcommune.web.controller.AvatarControllerTest.zip)

ctapobep:
From [what I found](http://javatalks.ru/topics/48287?page=1#245580) the version of Java and JodaTime may be inconsistent as they both have own DB of the timezones. What's your Java version? I'm using jdk1.8.0_60, maybe this is the problematic place.

vps2:
I have installed java jdk1.8.0_25.

ctapobep:
@vps2 would you be able to fix it yourself? I know that we can fix it just by using either JodaTime in the controllers or by using standard Java Date in the tests. But I don't want to do a blind fix without knowing what the problem is. And I cannot fix it on my own (neither our other devs) since it doesn't reproduce on my machine.

vps2:
Okay, I'll try to deal with this myself.

vps2:
After debugging the test, I found out that the error I have occurs when parsing the string representation of a date. If in the template the output format of the time zone change from "z" to "Z", the test begins to work.

test

From the picture we can see that in the string representation of a zone of a class java.util.Date has the inscription "GMT", and in jodatime DateTime - no.

ctapobep:
Browsers probably expect it in z form. I've got 1.8 and the same version of JodaTime and in my case it shows: `Mon, 18 Jan 2016 09:31:22 MSK`. Why...

vps2:
There may be the influence of regional settings in Windows. I have Windows 7 Pro SP1 x32 English with Formats and Location -> Russian.

ctapobep:
Would you be able to find out the exact way these dates are forming? I'd like to fix it once and for all, but I don't want to do this without knowing the exact reasons.

vps2:
I think my question can be removed. After updating java to the latest version it worked. 
My mistake! :)

ctapobep:
Well, we support Java 7+, so it should've worked on your machine as well..

PR #169, opened 2016-01-21
JC-2281 Error "User not found" error when the recipient is selected from list by the mouse

vps2:
- deleted "focusout" event handler of the fields "recipient", which is triggered before "select" events of the autocomplete widget. Instead, I added an event handler "change" to the autocomplete widget, that is triggered when the field is blurred, if the value has changed.


vps2:
Please tell me someone is addressing the issue JC-2282? For testing purposes I created a filter to solve this problem, but don't create a pullrequest.

ctapobep:
If you want to take it, you could write in comments of that ticket. I think @NCNecros hasn't started it yet.

NCNecros:
No, I did. We discuss tonight.

PR #170, opened 2016-01-24
JC-1090 Delete unnecessary URL when viewing image

res5515:
JC-1090 Delete unnecessary URL when viewing image

1. When user clicks on image the "PrettyPhoto" lib can either make new URL or not.
This parameter is configurable.

2.when user want to get back on the page, current "PrettyPhoto" lib get incoorect result.
 hashtag = (url.indexOf('#!prettyPhoto')) ? true : false; - always return true.
For resolving this problem I install newest version of the lib.


ctapobep:
There is no ticket id in the commit message. @res5515 could you please update it?

ctapobep:
Merged!

PR #172, opened 2016-02-15
JC-2274 Fixed bug: validation message about field length in drafts after changing focus doesn't disappear after deleting extra symbol

skythet:
Bug reason: function "wasChanged" compares the value of the field with the stored draft values rather than actual, so validation does not occur when deleting extra symbol

Solution: listen "input" event instead "onKeyUp" event, because "input" event triggered only if field changed


ctapobep:
Seems to be working, thanks! I'll merge it right after the release.

PR #173, opened 2016-02-16
JC-2285 Fixed bug: Text of editing comment in QnA topics appears after cancelling edit and trying re-edit

skythet:
Bug reason: confusion with "val()" and "text()" of textarea element, when cancel editing "val" stay changed

Solution: reset "val()" to "text()" when canceling edit and update "text()" when save comment


ctapobep:
Merged! Thanks.

PR #174, opened 2016-02-17
JC-2290 Fixed bug: Shift of search field when trying to type in it

skythet:
Instead of change display "search-clear" element, change "visibility", then element will keep place and input field will not jump.


ctapobep:
Merged!

PR #177, opened 2016-02-22
JC-2290 Fixing bug: Similar name of buttons in preview of editing answer

skythet:
Call button "Save answer" in a both modes (utf8 symbols)


ctapobep:
Looks like the commit message contains wrong bug number. Do you remember what was it?

ctapobep:
Found, it's JC-2286

PR #182, opened 2016-03-01
JC-2170 Wrong tags for sintax highlight in Opera12

Sabfir:
Some browsers could have selectionStart and selectionEnd that is not equal
to 0 for empty textarea.
E.g. opera 12 has 15. So it's better to assign 0 in order not to have
problems in the future.
I also incidentally found and corrected an inaccuracy in the file
bbeditor.tag.
It was
   `title='<spring:message code='label.answer.highlight'/>'>`
changed to
   `title="<spring:message code='label.answer.highlight'/>">`

### DIFF
--- a/jcommune-view/jcommune-web-view/src/main/webapp/resources/javascript/lib/wysiwyg-bbcode/editor.js
+++ b/jcommune-view/jcommune-web-view/src/main/webapp/resources/javascript/lib/wysiwyg-bbcode/editor.js
@@ -468,10 +468,11 @@ function addTag(t1, t2, selection) {
         sel_start = element.selectionStart;
         sel_end = element.selectionEnd;
 
-        if (element.value == "" && $.browser.opera) {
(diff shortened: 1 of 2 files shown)


jcommune-view/jcommune-web-view/src/main/webapp/resources/javascript/lib/wysiwyg-bbcode/editor.js
```diff
@@ -468,10 +468,11 @@ function addTag(t1, t2, selection) {
         sel_start = element.selectionStart;
         sel_end = element.selectionEnd;
 
-        if (element.value == "" && $.browser.opera) {
-            // for Opera browser null value (textarea empty) for selectionStart and selectionEnd is '20'
-            sel_start = sel_start - 20;
-            sel_end = sel_end - 20;
+        if (element.value == "") {
```
ctapobep: So it's a bug in the library?

PR #188, opened 2016-03-16
JC-2292 bug fixes

skythet:
Fixed next bugs:
- Groups are not sorted via alphabet
- Group list is blinking after first opening
- Groups multi select box is not translated on Russian and Ukrainian.
-Tab key does not move focus to the Groups button
- Network connection alert message has an html number of a single quote instead of a single quote symbol


jcommune-view/jcommune-web-view/src/main/webapp/WEB-INF/template/jsMessages.jsp
```diff
@@ -183,6 +183,6 @@ $labelNotLoggedInError = '<spring:message code="label.not.logged.in.error" htmlE
 $pollItemsSize = '<spring:message code="poll.items.size" htmlEscape="true" text=""/>';
 $pollItemLength = '<spring:message code="poll.item.length" htmlEscape="true" text=""/>';
 
-$labelConnectionLostGenericError = '<spring:message code="label.connection.lost.genericError" htmlEscape="true"/>';
+$labelConnectionLostGenericError = "<spring:message code="label.connection.lost.genericError" htmlEscape="false"/>";
```
ctapobep: This would allow for HTML Injections, wouldn't it? If we change the message at some point in the future there is a risk that we break HTML.

skythet: then, maybe we should use "`" symbol for "З'єднання" word?

ctapobep: No, that's hackish. We've got this symbol in other places, how is it working there?

skythet: Key `label.announcement` (Помітити як об'яву) used in `topicForm.jsp`:
```
<spring:message code="label.announcement"/>
```
without escape

ctapobep: Okay then, let's leave it. I guess this issue relates to #20

PR #191, opened 2016-03-20
Bug fixes

skythet:
Fixed bugs:
- Invalid search alert message in Groups list is not translated in Ukrainian and Russian.
- A common error message for adding/deleting to/from a group without authorization is not shown in all languages


jcommune-view/jcommune-web-view/src/main/webapp/resources/javascript/app/userSearch.js
```diff
@@ -38,8 +38,8 @@ userSearch.connectionErrorCallback = function(xhr, status) {
         } else {
             userSearch.popup.show($labelConnectionLostGenericError);
         }
-    } else if (xhr.status == 403) {
-        userSearch.popup.show($labelNotLoggedInError);
+    } else if (xhr.status == 403 || xhr.status == 500 || xhr.status == 302) {
+        userSearch.popup.show($labelNotLoggedInGenericError);
```
ctapobep: The errors don't seem to match the statuses.. 403 is access denied - agree. But error 500 is not. Neither 302 (though I assume this is a redirect to the login page?) - this we can change back in the security config if you think it'd be better.

jcommune-view/jcommune-web-view/src/main/webapp/resources/javascript/app/userSearch.js
```diff
@@ -38,7 +38,7 @@ userSearch.connectionErrorCallback = function(xhr, status) {
         } else {
             userSearch.popup.show($labelConnectionLostGenericError);
         }
-    } else if (xhr.status == 403 || xhr.status == 500 || xhr.status == 302) {
```
ctapobep: So what would happen in case of error 500 coming from the server?

PR #196, opened 2016-04-21
JC-1962 Fix Unexpected moderators list moving

NCNecros:


ctapobep:
@NCNecros, will merge this right after we release (should happen tomorrow)

ctapobep:
Merged

PR #197, opened 2016-04-23
JC-2309 common forum settings

skythet:


ctapobep:
When I enter english letters into session timeout I get error 500:

```
org.springframework.http.converter.HttpMessageNotReadableException: Could not read JSON: Can not construct instance of int from String value '301234132423412341234123v': not a va
lid Integer value
 at [Source: org.apache.catalina.connector.CoyoteInputStream@7cd4604e; line: 1, column: 211] (through reference chain: org.jtalks.jcommune.model.entity.ComponentInformation["sess
ionTimeout"]); nested exception is org.codehaus.jackson.map.JsonMappingException: Can not construct instance of int from String value '301234132423412341234123v': not a valid Int
eger value
 at [Source: org.apache.catalina.connector.CoyoteInputStream@7cd4604e; line: 1, column: 211] (through reference chain: org.jtalks.jcommune.model.entity.ComponentInformation["sess
ionTimeout"])
        at org.springframework.http.converter.json.MappingJacksonHttpMessageConverter.readJavaType(MappingJacksonHttpMessageConverter.java:179)
        at org.springframework.http.converter.json.MappingJacksonHttpMessageConverter.read(MappingJacksonHttpMessageConverter.java:171)
        at org.springframework.web.servlet.mvc.method.annotation.AbstractMessageConverterMethodArgumentResolver.readWithMessageConverters(AbstractMessageConverterMethodArgumentRe
solver.java:135)
```

jcommune-model/src/main/java/org/jtalks/jcommune/model/validation/annotations/IntegerRange.java
```diff
@@ -0,0 +1,64 @@
...
+
+import static java.lang.annotation.ElementType.*;
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+@Constraint(validatedBy = {IntegerRangeValidator.class})
+@Target({ METHOD, FIELD, ANNOTATION_TYPE, CONSTRUCTOR, PARAMETER })
+@Retention(RUNTIME)
+@ReportAsSingleViolation
+@NotBlank
+public @interface IntegerRange {
```
ctapobep: Please add javadocs describing the behaviour of this annotation.

jcommune-model/src/test/java/org/jtalks/jcommune/model/validation/IntegerRangeValidatorTest.java
```diff
@@ -0,0 +1,92 @@
...
+    @BeforeMethod
+    public void init() {
+        initMocks(this);
+    }
+
+    @Test
+    public void testValidate() {
+        IntegerRangeValidator validator = new IntegerRangeValidator();
+
+        when(integerRange.min()).thenReturn(1);
```
ctapobep: You can create method integerRange(int min, int max) that create the range and doesn't pollute the test with unnecessary code.

jcommune-model/src/test/java/org/jtalks/jcommune/model/validation/IntegerRangeValidatorTest.java
```diff
@@ -0,0 +1,92 @@
...
+        when(integerRange.max()).thenReturn(15);
```
ctapobep: Also, instead of using concrete integers let's use Datagen to generate random values: https://bitbucket.org/qala/datagen/

jcommune-model/src/test/java/org/jtalks/jcommune/model/validation/IntegerRangeValidatorTest.java
```diff
@@ -0,0 +1,92 @@
...
+
+import static org.mockito.Mockito.when;
+import static org.mockito.MockitoAnnotations.initMocks;
+import static org.testng.Assert.assertFalse;
+import static org.testng.Assert.assertTrue;
+
+/**
+ * @author skythet
+ */
+public class IntegerRangeValidatorTest {
```
ctapobep: The test cases missing:
- With min/max lower than 0
- With min/max equal to each other
- With min > max

ctapobep:
If errors don't appear anymore, let's squash the commits into 1 and merge into master. And finish writing tests in parallel with QA team looking at the story.

pom.xml
```diff
@@ -471,6 +471,11 @@
         <version>4.11</version>
         <scope>test</scope>
       </dependency>
+      <dependency>
+        <groupId>io.qala.datagen</groupId>
+        <artifactId>qala-datagen</artifactId>
+        <version>1.2.1</version>
```
ctapobep: scope=test as for now will suffice.

ctapobep:
Merged

PR #198, opened 2016-05-04
Fix the failed test

NCNecros:


jcommune-model/src/test/java/org/jtalks/jcommune/model/dao/hibernate/GroupHibernateDaoTest.java
```diff
@@ -101,6 +104,12 @@ public void testGetAll() {
         saveAndEvict(group1);
 
         List<Group> actual = groupDao.getAll();
+        Collections.sort(actual, new Comparator<Group>() {
```
ctapobep: To keep tests declarative, could you please move this into a separate method like `sortById(List<Group>)`?

ctapobep: And squash right after that - I'll merge it to master

NCNecros: ok. i did it.

PR #201, opened 2016-05-18
JC-2327 Admin popup: Removal of Logo/Fav Icon cancellation leads to Error 500

Noctrunal:


jcommune-view/jcommune-web-view/src/main/webapp/resources/javascript/app/forumAdministration.js
```diff
@@ -427,6 +427,11 @@ function addRestoreDefaultImageHandler(buttonId, defaultImageUrl, onSuccess) {
             });
             jDialog.closeDialog();
         };
+        
+        var cancelFunc = function () {
```
ctapobep: No need to add Func - it's clear that this is a function

ctapobep:
Could you please rephrase the commit message and include explanation of why the error was happening and how your changes fix it?

PR #206, opened 2016-05-23

mihnayan:

- added the path for `contextFolder` parameter specified in Windows style
(with back slashes) to pom.xml:
```xml
<contextFolder>${basedir}/src/main/webapp/,${basedir}\src\main\webapp\</contextFolder>
```

described problem.

### DIFF
--- a/jcommune-view/jcommune-web-view/pom.xml
+++ b/jcommune-view/jcommune-web-view/pom.xml
@@ -142,7 +142,7 @@
       <plugin>
         <groupId>ro.isdc.wro4j</groupId>
         <artifactId>wro4j-maven-plugin</artifactId>
-        <version>1.7.5</version>
+        <version>1.7.6</version>
         <executions>
           <execution>
             <phase>generate-resources</phase>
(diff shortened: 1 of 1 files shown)


jcommune-view/jcommune-web-view/pom.xml
```diff
@@ -157,7 +157,7 @@
           <destinationFolder>${basedir}/src/main/webapp/resources/wro</destinationFolder>
           <cssDestinationFolder>${basedir}/src/main/webapp/resources/wro</cssDestinationFolder>
           <jsDestinationFolder>${basedir}/src/main/webapp/resources/wro</jsDestinationFolder>
-          <contextFolder>${basedir}/src/main/webapp/</contextFolder>
+          <contextFolder>${basedir}/src/main/webapp/,${basedir}\src\main\webapp\</contextFolder>
```
ctapobep: Just one small thing - could copy-paste a part of your commit message as a comment to this line? People won't see the commit message unless they search for it.

PR #207, opened 2016-05-23

mihnayan:

- in pom.xml added additional path for `contextFolder` parameter specified
in Windows style (with back slashes):
```xml
<contextFolder>${basedir}/src/main/webapp/,${basedir}\src\main\webapp\</contextFolder>
```

For more information about `contextFolder` parameter see plugin's official
documentation: https://github.com/wro4j/wro4j/blob/1.7.x/docs/MavenPlugin.md#plugin-parameters

described problem.

### DIFF
--- a/jcommune-view/jcommune-web-view/pom.xml
+++ b/jcommune-view/jcommune-web-view/pom.xml
@@ -142,7 +142,7 @@
       <plugin>
         <groupId>ro.isdc.wro4j</groupId>
         <artifactId>wro4j-maven-plugin</artifactId>
-        <version>1.7.5</version>
+        <version>1.7.6</version>
         <executions>
           <execution>
             <phase>generate-resources</phase>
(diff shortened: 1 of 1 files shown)


jcommune-view/jcommune-web-view/pom.xml
```diff
@@ -157,7 +157,7 @@
           <destinationFolder>${basedir}/src/main/webapp/resources/wro</destinationFolder>
           <cssDestinationFolder>${basedir}/src/main/webapp/resources/wro</cssDestinationFolder>
           <jsDestinationFolder>${basedir}/src/main/webapp/resources/wro</jsDestinationFolder>
-          <contextFolder>${basedir}/src/main/webapp/</contextFolder>
+          <contextFolder>${basedir}/src/main/webapp/,${basedir}\src\main\webapp\</contextFolder>
```
ctapobep: Looks like you updated only the commit message :)

mihnayan: Sorry, I probably have misunderstood. Should I paste a comments in the pom.xml?

ctapobep: Yep :)

PR #208, opened 2016-05-24

mihnayan:

- in pom.xml added additional path for `contextFolder` parameter specified
in Windows style (with back slashes).

described problem.


ctapobep:
We did it :) Thanks a lot for the research.

PR #216, opened 2016-06-08
JC-2326 Alert message 'Access denied' bug fix

Noctrunal:
Removed wrong intercept-url for path /admin/edit
because if it there Controller can't get request and
handle AccessDeniedException for user in jQuery
alert dialog.


jcommune-view/jcommune-web-view/src/main/webapp/resources/javascript/app/forumAdministration.js
```diff
@@ -538,7 +538,7 @@ function sendForumConfiguration(e) {
                 } else {
                     jDialog.createDialog({
                         type: jDialog.alertType,
-                        bodyMessage: resp.result
+                        bodyMessage: $labelAccessDenied
```
ctapobep: Don't you want to check for the error code? Currently if any error happens, you'd show access denied.

ctapobep:
Merged!

PR #217, opened 2016-06-13
JC-2289 XSS vulnerability at profile page bug fix

Noctrunal:
Included JSTL's c:out tag for
${editedUser.username} expression
in editProfile.jsp. Because this tag
escapes HTML characters for
avoid cross-site scripting.


ctapobep:
This doesn't seem to fix the issue, try `'><div>`. Note, that in your commit message or in comments in the code you need to add the proof that this change really fixes something. Do you have a proof that adding quotes fixes XSS?

PR #224, opened 2016-08-11
JC-2049 Fixed "Sign In" middle mouse button click

lekarto:
signin function was replaced from anonymous to separate function because it's a bad practice to write function inside the general initialization method.


ctapobep:
Merged

PR #226, opened 2016-08-18
Update version of Flyway to 2.3.1

mihnayan:
Interim update of the Flyway for subsequent updates to version 4.0.
Optimized Flyway beans in application Context.

Change version of HSQLDB to 2.3.3 for test usage.
Used explicit type conversion in Topic and Branch entities because
auto casting behaviour has become more strict in HSQLDB.


ctapobep:
Could you please rebase on master?

PR #227, opened 2016-08-18
JC-2273 Map files are generated when JS compression happens

MuH3gPaB:
Wro4j doesn't support SourceMaps, in this case another stuff was used.
Now js and css files are compressed by primefaces extensions
resources optimizer maven plugin.
Sourcemaps are generated at the same moment.

All configuration in jcommunue-web-view/pom.xml
wro.xml is not necessary any more.

Next step: Reorganize resources folder, so we could  make pom.xml more compact.


ctapobep:
Looks good at first glance. From what I understood you kept the same paths/filenames but just used a different plugin, right? 

@masyan, what do you think?

ctapobep:
Did you check that there are no warnings/errors in the JS console?

masyan:
'Wro4j doesn't support SourceMaps' you mean wildcard? wro4j supprt id (http://wro4j.readthedocs.io/en/stable/WildcardSupport/). So, size will not decrease.

MuH3gPaB:
> you kept the same paths/filenames but just used a different plugin, right?

yes, you right
i've used another plugin, that supports generating source maps
all paths/filenames don't changed
I don't think paths are good for this config, cause we have to aggregate 
files from different subfolders, so we have to include each file separately
but it's not a problem, everything work fine

> Did you check that there are no warnings/errors in the JS console?

I think there was no errors or warnings, I'll check again tomorrow

> 'Wro4j doesn't support SourceMaps' you mean wildcard?

Not right. SourceMaps used to restore original sources from compressed js

MuH3gPaB:
> Did you check that there are no warnings/errors in the JS console?

In JS console are two warnings same as at javatalks.ru

Lots of warnings at CSS console, and they are same as at javatalks.ru too.

ctapobep:
Awesome. Will merge this when the new spring starts.

jcommune-view/jcommune-web-view/pom.xml
```diff
@@ -140,35 +140,290 @@
...
+                <!--LIBS - CSS-->
+                <include>/resources/css/lib/jquery.contextMenu.css</include>
+                <include>/resources/css/lib/jquery-ui.css</include>
+                <include>/resources/css/lib/chosen.css</include>
+              </includes>
+
+              <aggregations>
+                <aggregation>
+                  <outputFile>
+                    ${basedir}/src/main/webapp/resources/wro/main.css
```
masyan: remove "wro"

jcommune-view/jcommune-web-view/pom.xml
```diff
@@ -140,35 +140,290 @@
...
+                <include>resources/javascript/app/events.js</include>
+                <include>resources/javascript/app/banner.js</include>
+                <include>resources/javascript/app/forumAdministration.js</include>
+                <include>resources/javascript/app/editPermissions.js</include>
+              </includes>
+
+              <aggregations>
+                <aggregation>
+                  <outputFile>
+                    ${basedir}/src/main/webapp/resources/wro/main.js
```
masyan: wro

jcommune-view/jcommune-web-view/pom.xml
```diff
@@ -140,35 +140,290 @@
...
+                <include>resources/javascript/app/contextMenu.js</include>
+                <include>resources/javascript/lib/purl.js</include>
+                <include>resources/javascript/app/pollPreview.js</include>
+                <include>resources/javascript/app/codeHighlighting.js</include>
+              </includes>
+
+              <aggregations>
+                <aggregation>
+                  <outputFile>
+                    ${basedir}/src/main/webapp/resources/wro/pm.js
```
masyan: etc.

ctapobep:
@MuH3gPaB could you please squash everything into 1 commit? 

Also, if it's not hard could you please try running system functional tests (via selenium) locally (since there are a lot of changes that potentially could've broken any place in JC): https://github.com/jtalks-org/functional-tests/tree/master/functional-tests-jcommune

They take a lot of time (an hour or two), so it's worth to leave them running when you go to bed :)

MuH3gPaB:
Ok. I'll do it.

MuH3gPaB:
Commits squashed, tests passed =) everything looks fine.

ctapobep:
Merged!

PR #237, opened 2016-09-11
DELETED

egus032:
Please, use commit with number 2f30b37


jcommune-view/jcommune-web-view/src/main/webapp/WEB-INF/tags/topicList.tag
```diff
@@ -28,7 +28,7 @@
             <jtalks:topicIconSmall topicDto="${topicDto}" authenticated="${authenticated}"/>
           </td>
           <td class="posts-td-small posts-td-small_2">
-            <h2 class="h-nostyle">
+            <h2 class="h-nostyle"  hNoStyleId="${topicDto.topic.id}">
```
ctapobep: No need for this one.

PR #240, opened 2016-09-20
JC-2350 FIX invalid connection encoding

NCNecros:


oatkachenko:
We have removed characterEncoding parameter from jdbc connection url because since version 5.1.13 Connector/J did not support utf8mb4 for servers 5.5.2 and newer. Now parameter is taken from the server settings (https://dev.mysql.com/doc/relnotes/connector-j/5.1/en/news-5-1-13.html)

NCNecros:
Probably it didn't took the parameter from server. The characterEncoding fix errors with rombus instead utf-8 characters.

ctapobep:
@NCNecros can we find a different solution to this? We want to use the server encoding for sure..

PR #241, opened 2016-09-20
JC-2366 Error 500 displayed instead of 405

oatkachenko:
In case when method in request can't be applied to requested resource we want to show proper error page.
- added property in PrettyLogExceptionResolver bean to catch HttpRequestMethodNotSupportedException
- added 405 error page
- added error messages and corresponding translation
- added request mapping for 405 error


ctapobep:
@oatkachenko  the commit message doesn't reference the JIRA ticket. Can we add the reference?

ctapobep:
Merged!

PR #244, opened 2016-09-28
Added basic Gatling tests

Varro88:
Added login, branch, topic and recent activity tests via Gatling

### DIFF
--- a/jcommune-performance-tests/src/test/scala/org/jtalks/jcommune/performance/tests/Login.scala
+++ b/jcommune-performance-tests/src/test/scala/org/jtalks/jcommune/performance/tests/Login.scala
@@ -1,0 +1,44 @@
+/**
+  * Copyright (C) 2011  JTalks.org Team
+  * This library is free software; you can redistribute it and/or
+  * modify it under the terms of the GNU Lesser General Public
+  * License as published by the Free Software Foundation; either
+  * version 2.1 of the License, or (at your option) any later version.
+  * This library is distributed in the hope that it will be useful,
+  * but WITHOUT ANY WARRANTY; without even the implied warranty of
+  * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+  * Lesser General Public License for more details.
+  * You should have received a copy of the GNU Lesser General Public
+  * License along with this library; if not, write to the Free Software
+  * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+  */
+package org.jtalks.jcommune.performance.tests
+
+import io.gatling.core.Predef._
+import io.gatling.http.Predef._
+
+class Login extends Simulation{
+  val url = "http://performance.jtalks.org/jcommune"
+  val username = "admin"
+  val password = "admin"
+
+  val httpConf = http
+    .baseURL(url)
+    .acceptHeader("text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8")
+    .doNotTrackHeader("1")
+    .acceptLanguageHeader("en-US,en;q=0.5")
+    .acceptEncodingHeader("gzip, deflate")
+    .userAgentHeader("Mozilla/5.0 (Windows NT 5.1; rv:31.0) Gecko/20100101 Firefox/31.0")
+
+  val scn = scenario("BasicSimulation")
+    .exec(http("request_1")
+      .post("/login_ajax")
+      .formParam("userName", username)
+      .formParam("password", password)
+      .formParam("_spring_security_remember_me", "on"))
+    .pause(5)
+
+  setUp(
+    scn.inject(constantUsersPerSec(3) during DurationInteger(1800).seconds randomized)
+  ).protocols(httpConf)
+}
--- a/jcommune-performance-tests/src/test/scala/org/jtalks/jcommune/performance/tests/OpenBranchPage.scala
+++ b/jcommune-performance-tests/src/test/scala/org/jtalks/jcommune/performance/tests/OpenBranchPage.scala
@@ -1,0 +1,40 @@
+/**
+  * Copyright (C) 2011  JTalks.org Team
+  * This library is free software; you can redistribute it and/or
+  * modify it under the terms of the GNU Lesser General Public
+  * License as published by the Free Software Foundation; either
+  * version 2.1 of the License, or (at your option) any later version.
+  * This library is distributed in the hope that it will be useful,
+  * but WITHOUT ANY WARRANTY; without even the implied warranty of
+  * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+  * Lesser General Public License for more details.
+  * You should have received a copy of the GNU Lesser General Public
+  * License along with this library; if not, write to the Free Software
+  * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+  */
+package org.jtalks.jcommune.performance.tests
+
+import io.gatling.core.Predef._
+import io.gatling.http.Predef._
+
+class OpenBranchPage extends Simulation{
+  val url = "http://performance.jtalks.org/jcommune"
+
+  val httpConf = http
+    .baseURL(url)
+    .acceptHeader("text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8")
+    .doNotTrackHeader("1")
+    .acceptLanguageHeader("en-US,en;q=0.5")
+    .acceptEncodingHeader("gzip, deflate")
+    .userAgentHeader("Mozilla/5.0 (Windows NT 5.1; rv:31.0) Gecko/20100101 Firefox/31.0")
+
+  val scn = scenario("BasicSimulation")
+    .exec(http("request_1")
+      .get("/branches/1"))
+    .pause(5)
+
+  setUp(
+    scn.inject(constantUsersPerSec(30) during DurationInteger(1800).seconds randomized)
+  ).protocols(httpConf)
+}
+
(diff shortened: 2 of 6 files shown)


jcommune-performance-tests/src/test/scala/org/jtalks/jcommune/performance/tests/Login.scala
```diff
@@ -0,0 +1,44 @@
...
+  * License along with this library; if not, write to the Free Software
+  * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+  */
+package org.jtalks.jcommune.performance.tests
+
+import io.gatling.core.Predef._
+import io.gatling.http.Predef._
+
+class Login extends Simulation{
+  val url = "http://performance.jtalks.org/jcommune"
```
oatkachenko: I think that all constants shold be placed in another class (e.g. Utils) and passed as an arguments to JVM. In this case configuring the tests will be more simple.

jcommune-performance-tests/src/test/scala/org/jtalks/jcommune/performance/tests/OpenTopicPage.scala
```diff
@@ -0,0 +1,41 @@
...
+  * You should have received a copy of the GNU Lesser General Public
+  * License along with this library; if not, write to the Free Software
+  * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+  */
+package org.jtalks.jcommune.performance.tests
+
+import io.gatling.core.Predef._
+import io.gatling.http.Predef._
+
+class OpenForumMainPage extends Simulation{
```
oatkachenko: Filename and class name should be the same.

jcommune-performance-tests/src/test/scala/org/jtalks/jcommune/performance/tests/Login.scala
```diff
@@ -0,0 +1,44 @@
...
+  val username = "admin"
+  val password = "admin"
+
+  val httpConf = http
+    .baseURL(url)
+    .acceptHeader("text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8")
+    .doNotTrackHeader("1")
+    .acceptLanguageHeader("en-US,en;q=0.5")
+    .acceptEncodingHeader("gzip, deflate")
+    .userAgentHeader("Mozilla/5.0 (Windows NT 5.1; rv:31.0) Gecko/20100101 Firefox/31.0")
```
oatkachenko: val httpConf initialized the same way in all classes. I think it can be moved to another class (e.g. object Utils) and just call Utils.httpConf every time when it needed.

jcommune-performance-tests/src/test/scala/org/jtalks/jcommune/performance/tests/OpenBranchPage.scala
```diff
@@ -0,0 +1,40 @@
...
+    .baseURL(url)
+    .acceptHeader("text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8")
+    .doNotTrackHeader("1")
+    .acceptLanguageHeader("en-US,en;q=0.5")
+    .acceptEncodingHeader("gzip, deflate")
+    .userAgentHeader("Mozilla/5.0 (Windows NT 5.1; rv:31.0) Gecko/20100101 Firefox/31.0")
+
+  val scn = scenario("BasicSimulation")
+    .exec(http("request_1")
+      .get("/branches/1"))
```
oatkachenko: It is not good idea to open the same branch every time. I think that random page should be opened.

jcommune-performance-tests/src/test/scala/org/jtalks/jcommune/performance/tests/Login.scala
```diff
@@ -0,0 +1,44 @@
...
+  val scn = scenario("BasicSimulation")
+    .exec(http("request_1")
+      .post("/jcommune/login_ajax")
+      .formParam("userName", username)
+      .formParam("password", password)
+      .formParam("_spring_security_remember_me", "on"))
+    .pause(5)
+
+  setUp(
+    scn.inject(atOnceUsers(1))
```
oatkachenko: It will add only one user, but it is not enough. I think we should add more users during several minutes or longer.

jcommune-performance-tests/src/test/scala/org/jtalks/jcommune/performance/tests/OpenTopicPage.scala
```diff
@@ -0,0 +1,41 @@
...
+  val url = "http://performance.jtalks.org/jcommune"
+  val topicId = "11111"
```
oatkachenko: I think topicId can be chosen randomly.

jcommune-performance-tests/src/test/scala/org/jtalks/jcommune/performance/tests/Login.scala
Varro88: Should we use something like `constantUsersPerSec(rate) during(duration) randomized` with 20 users per second during 5 minutes for all tests?

jcommune-performance-tests/src/test/scala/org/jtalks/jcommune/performance/tests/OpenBranchPage.scala
Varro88: 1) Do we need random branch/topic per test or per user in test?
2) What is the best way to get valid branc/topic ids (to avoid opening non-existence elements)? Should we get it dynamically or have some predefined set?

jcommune-performance-tests/src/test/scala/org/jtalks/jcommune/performance/tests/Login.scala
oatkachenko: Yes, you are on the right way.

jcommune-performance-tests/src/test/scala/org/jtalks/jcommune/performance/tests/OpenBranchPage.scala
oatkachenko: 1) Random branch/topic for each user.
2) You can build Scenario with ChainBuilder for example:
Open branches page -> Extract all branche ids into List -> Open branch with id randomly picked from the List. At the end scenario should contain 3 Actions. Nearly the same steps for topics.

jcommune-performance-tests/src/test/scala/org/jtalks/jcommune/performance/tests/Login.scala
ctapobep: I agree. But this can be done a little bit later.

jcommune-performance-tests/src/test/scala/org/jtalks/jcommune/performance/tests/Login.scala
ctapobep: Same - let's do this with the next commit.

jcommune-performance-tests/src/test/scala/org/jtalks/jcommune/performance/tests/Login.scala
ctapobep: I think we had different plans for JMeter.. We could reuse numbers from there, couldn't we?

jcommune-performance-tests/src/test/scala/org/jtalks/jcommune/performance/tests/OpenBranchPage.scala
ctapobep: Let's deal with all the hardcodes in the next commits - right now our goal is to have something working and then we'll start improving it.

jcommune-performance-tests/src/test/scala/org/jtalks/jcommune/performance/tests/Login.scala
```diff
@@ -0,0 +1,44 @@
...
+    .baseURL(url)
+    .acceptHeader("text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8")
+    .doNotTrackHeader("1")
+    .acceptLanguageHeader("en-US,en;q=0.5")
+    .acceptEncodingHeader("gzip, deflate")
+    .userAgentHeader("Mozilla/5.0 (Windows NT 5.1; rv:31.0) Gecko/20100101 Firefox/31.0")
+
+  val scn = scenario("BasicSimulation")
+    .exec(http("request_1")
+      .post("/jcommune/login_ajax")
```
oatkachenko: It will not work because requested url will be "http://performance.jtalks.org/jcommune/jcommune/login_ajax"

jcommune-performance-tests/src/test/scala/org/jtalks/jcommune/performance/tests/Login.scala
```diff
@@ -0,0 +1,44 @@
...
+      .formParam("userName", username)
+      .formParam("password", password)
+      .formParam("_spring_security_remember_me", "on"))
+    .pause(5)
+
+  setUp(
+    scn.inject(atOnceUsers(30))
```
oatkachenko: Could you please change it to scn.inject(constantUsersPerSec(3) during DurationInteger(1800).seconds randomized) because we used this values in jMeter.

jcommune-performance-tests/src/test/scala/org/jtalks/jcommune/performance/tests/OpenBranchPage.scala
```diff
@@ -0,0 +1,40 @@
...
+    .pause(5)
+
+  setUp(
+    scn.inject(atOnceUsers(30))
```
oatkachenko: The same, injection should be like (constantUsersPerSec(30) during DurationInteger(1800).seconds randomized), but this time, as you can see, 30 users. We must generate 3 registered users and 30 anonymous during 1800 seconds.

ctapobep:
Merged!

PR #251, opened 2016-11-07
JC-2360 Speed Up Forum

oatkachenko:
- updated jtalks-common version to 0.62
- decreased number of sql queries generated in hasPermission() method.

### DIFF
--- a/jcommune-service/src/main/java/org/jtalks/jcommune/service/security/AclGroupPermissionEvaluator.java
+++ b/jcommune-service/src/main/java/org/jtalks/jcommune/service/security/AclGroupPermissionEvaluator.java
@@ -21,13 +21,10 @@ import java.util.List;
 import javax.annotation.Nonnull;
 
 import org.apache.commons.lang.Validate;
-//import org.jtalks.common.model.dao.GroupDao;
-//import org.jtalks.common.model.entity.Group;
-//import org.jtalks.common.model.entity.User;
-import org.jtalks.common.model.dao.GroupDao;
 import org.jtalks.common.model.entity.Group;
 import org.jtalks.common.model.permissions.BranchPermission;
 import org.jtalks.common.model.permissions.GeneralPermission;
+import org.jtalks.common.model.permissions.JtalksPermission;
 import org.jtalks.common.model.permissions.ProfilePermission;
 import org.jtalks.common.security.acl.AclManager;
 import org.jtalks.common.security.acl.AclUtil;
@@ -65,7 +62,6 @@ public class AclGroupPermissionEvaluator implements PermissionEvaluator {
 
     private final AclManager aclManager;
     private final AclUtil aclUtil;
-    private final GroupDao groupDao;
     private final JtalksSidFactory sidFactory;
     private final JdbcMutableAclService mutableAclService;
     private final UserDao userDao;
@@ -74,13 +70,11 @@ public class AclGroupPermissionEvaluator implements PermissionEvaluator {
     /**
      * @param aclManager        for getting permissions on object indentity
      * @param aclUtil           utilities to work with Spring ACL
-     * @param groupDao          dao for user group getting
      * @param sidFactory        factory to work with principals
      * @param mutableAclService for checking existing of sids
      */
     public AclGroupPermissionEvaluator(@Nonnull org.jtalks.common.security.acl.AclManager aclManager,
                                        @Nonnull AclUtil aclUtil,
-                                       @Nonnull GroupDao groupDao,
                                        @Nonnull JtalksSidFactory sidFactory,
                                        @Nonnull JdbcMutableAclService mutableAclService,
                                        @Nonnull UserDao userDao,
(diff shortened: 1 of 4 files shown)


jcommune-service/src/main/java/org/jtalks/jcommune/service/security/AclGroupPermissionEvaluator.java
```diff
@@ -16,10 +16,12 @@
...
 import java.io.Serializable;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 
 import javax.annotation.Nonnull;
 
+import org.apache.commons.collections.CollectionUtils;
 import org.apache.commons.lang.Validate;
 //import org.jtalks.common.model.dao.GroupDao;
```
ctapobep: Could you please clean up these comments?

ctapobep: Also, `groupDao` field seems to be unused anymore.

jcommune-service/src/main/java/org/jtalks/jcommune/service/security/AclGroupPermissionEvaluator.java
```diff
@@ -113,33 +116,30 @@ public boolean hasPermission(Authentication authentication, Object targetId, Obj
...
         }
 
-        if (permission == ProfilePermission.EDIT_OWN_PROFILE &&
-                ((JCUser) authentication.getPrincipal()).getId() != id) {
-            return false;
-        }
-
         if (isRestrictedForSid(sid, aces, jtalksPermission) ||
                 isRestrictedForGroup(controlEntries, authentication, jtalksPermission) ||
                 isRestrictedPersonalPermission(authentication, jtalksPermission)) {
```
ctapobep: I thought we were going to check if the permission is a personal permission before checking the rest.. 

Though I see a check below. But I think that if it _is_ a PersonalPermission, then it doesn't make sense to check others. So it makes sense to put that check here.

jcommune-service/src/main/java/org/jtalks/jcommune/service/security/AclGroupPermissionEvaluator.java
```diff
@@ -340,10 +341,14 @@ private boolean isRestrictedForGroup(List<GroupAce> controlEntries,
     private boolean isGrantedForGroup(List<GroupAce> controlEntries,
                                       Authentication authentication, Permission permission,
                                       boolean isCheckAllowedGrant) {
-        if (authentication.getPrincipal() instanceof JCUser) {
+        Object auth = authentication.getPrincipal();
+        if (auth instanceof JCUser) {
```
ctapobep: Too nested.. Can be simplified:

```java
        Object auth = authentication.getPrincipal();
        if (!(auth instanceof JCUser)) return false;
        
        List<Long> groupsIDs = ((JCUser) auth).getGroupsIDs();
        for (GroupAce ace : controlEntries) {
            if (groupsIDs.contains(ace.getGroupId())) {
                if (isGrantedForGroup(ace, permission, isCheckAllowedGrant)) return true;
            }
        }
        return false;
```

jcommune-service/src/main/java/org/jtalks/jcommune/service/security/AclGroupPermissionEvaluator.java
```diff
@@ -340,10 +341,14 @@ private boolean isRestrictedForGroup(List<GroupAce> controlEntries,
...
+            List<Long> groupsIDs = ((JCUser) auth).getGroupsIDs();
```
ctapobep: Hypothetically (!) can be improved with a HashSet since its complexity is O(1) instead of O(n). But that depends on the size of `groupIDs` and `controlEntries` as if they are small then building a HashSet is probably going to take more time than to iterate over the arrays multiple times.

oatkachenko: I also thought to use HashSet here, but 99% of users are members of only one group, and controlEntries size not so large because it has been filtered previously.

jcommune-service/src/main/java/org/jtalks/jcommune/service/security/AclGroupPermissionEvaluator.java
```diff
@@ -113,43 +106,38 @@ public boolean hasPermission(Authentication authentication, Object targetId, Obj
...
         } catch (NotFoundException nfe) {
             aces = new ArrayList<>();
             controlEntries = new ArrayList<>();
         }
-
-        if (permission == ProfilePermission.EDIT_OWN_PROFILE &&
-                ((JCUser) authentication.getPrincipal()).getId() != id) {
-            return false;
+        if (jtalksPermission instanceof ProfilePermission
+                && authentication.getPrincipal() instanceof JCUser){
```
ctapobep: Could you please put both conditions on 1 line? it'll be easier to read..

PR #255, opened 2016-11-25
JC-2367 Edit user groups

oatkachenko:
### Changes
- changed 'groupCreateDialog', now it allows you to edit or create group.
- changed request mapping for 'create new group' in 'AdministrationController' to follow REST principles.
- added translations
- added css styles for groupAdministration page
- 'create group' performs 'POST' request on /group
- 'edit group' performs 'PUT' request on /group/{id}

### Next steps
- realize method isEditable() //returns false if group.name is equal to one of the next group names: {"Administrators", "Registered Users", "Banned Users"}, otherwise  returns true.
- add constructor(long id, String name, String description, int count) to populate DTO right from HQL query.

Since GroupAdministrationDto was changed we should change all tests connected with it


jcommune-view/jcommune-web-view/src/main/webapp/resources/javascript/app/forumAdministration.js
```diff
@@ -558,13 +565,22 @@ function sendForumConfiguration(e) {
...
+function showGroupManagementDialog(event) {
     event.preventDefault();
-
+    // Create a new group or edit an existing?
+    var editMode = $(this).attr('id') == "editGroup";
+
+    if (editMode){
+        // find row with group and extract all data that we need.
+        var groupRow = $(this).parents('.group-edit-row');
+        var groupId = groupRow.attr('id').substring('group-id'.length);
```
ctapobep: I'd rather create a hidden field or another attribute to eliminate additional logic with strings.

jcommune-view/jcommune-web-view/src/main/webapp/resources/javascript/app/forumAdministration.js
```diff
@@ -590,24 +606,27 @@ function showGroupCreateDialog(event) {
...
+        groupInformation.id = groupId;
         groupInformation.name = jDialog.dialog.find('#groupName').val();
         groupInformation.description = jDialog.dialog.find('#groupDescription').val();
 
         jDialog.dialog.find('*').attr('disabled', true);
 
         $.ajax({
-            url: $root + '/group/new',
-            type: 'POST',
+            url: $root + '/group/' + (editMode ? groupId : 'new'),
```
ctapobep: No "new" part is needed. Or is it for backward compatibility until @Temdegon changes the controllers?

jcommune-view/jcommune-web-view/src/main/webapp/resources/javascript/app/forumAdministration.js
```diff
@@ -634,4 +653,13 @@ function showGroupCreateDialog(event) {
             }
         });
     }
+
+    /**
+     *  When groupManagementDialog in 'edit group' mode this method fills up
+     *  the fields in this dialog.
+     */
+    function fillGroupManagementDialogFields() {
```
ctapobep: Can you instead pass the values as the parmaterers? Instead of relying on global state and the order of execution.

jcommune-view/jcommune-web-view/src/main/webapp/WEB-INF/jsp/groupAdministration.jsp
```diff
@@ -42,9 +42,16 @@
         </tr>
         </thead>
         <c:forEach var="group" items="${groups}">
-            <tr class="grid-row">
-                <td><c:out value="${group.name}"/></td>
-                <td><c:out value="${group.numberOfUsers}"/></td>
+            <tr id='group-id${group.id}' class="grid-row group-edit-row">
+                <td id="group-name"><c:out value="${group.name}"/></td>
```
ctapobep: I think we intend with 2 symbols in XML-like structures.

oatkachenko:
fixed.

jcommune-view/jcommune-web-view/src/main/webapp/resources/javascript/app/forumAdministration.js
```diff
@@ -590,24 +606,27 @@ function showGroupCreateDialog(event) {
...
+        groupInformation.id = groupId;
         groupInformation.name = jDialog.dialog.find('#groupName').val();
         groupInformation.description = jDialog.dialog.find('#groupDescription').val();
 
         jDialog.dialog.find('*').attr('disabled', true);
 
         $.ajax({
-            url: $root + '/group/new',
-            type: 'POST',
+            url: $root + '/group/' + (editMode ? groupId : ''),
```
ctapobep: So if we merge this right now the page won't function anymore?

oatkachenko: yes it won't work until we change the controller

ctapobep: So let's then leave your previous version here and then @Temdegon will update this logic when he refactors the Controllers. Or you can do the refactoring right in this commit since it doesn't relate to the new logic.

oatkachenko: I've made changes in DTO and Controller (as for new group creation). Now 'group administration' page works properly. We can add the group, but can't edit (because they are not editable) till @Temdegon refactor the Controller.

jcommune-model/src/main/java/org/jtalks/jcommune/model/dto/GroupAdministrationDto.java
```diff
@@ -67,4 +77,10 @@ public void setNumberOfUsers(long numberOfUsers) {
         this.numberOfUsers = numberOfUsers;
     }
 
+    public boolean isEditable() {
+        //TODO
+        // Implement logic that evaluates is group editable. Something like this:
+        // Arrays.asList("Administrators", "Registered Users", "Banned Users").contains(name)
+        return editable;
```
ctapobep: This is a good example of how TDD could work - you could already write a unit test for this.

ctapobep:
Merged

PR #263, opened 2016-12-22
JC-2380 SpeedUpForum

oatkachenko:
- Everytime when we need to get the current user, we searched it by name (case insensitive way). So indexes wasn't worked and we searched through all users. Now current user is taken from session.

### DIFF
--- a/jcommune-service/src/main/java/org/jtalks/jcommune/service/transactional/TransactionalUserService.java
+++ b/jcommune-service/src/main/java/org/jtalks/jcommune/service/transactional/TransactionalUserService.java
@@ -18,9 +18,9 @@ import org.apache.commons.lang.RandomStringUtils;
 import org.joda.time.DateTime;
 import org.joda.time.Period;
 import org.jtalks.common.model.dao.GroupDao;
-import org.jtalks.common.model.entity.Group;
 import org.jtalks.common.model.entity.User;
 import org.jtalks.common.security.SecurityService;
+import org.jtalks.common.service.security.SecurityContextFacade;
 import org.jtalks.jcommune.model.dao.PostDao;
 import org.jtalks.jcommune.model.dao.UserDao;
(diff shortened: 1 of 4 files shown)


jcommune-model/src/main/resources/org/jtalks/jcommune/model/entity/JCUser.hbm.xml
```diff
@@ -65,7 +65,7 @@
                 type="org.joda.time.contrib.hibernate.PersistentDateTime" not-null="false"/>
       <property column="POST_COUNT" name="postCount"/>
 
-      <set name="contacts" cascade="all-delete-orphan" inverse="true">
+      <set name="contacts" cascade="all-delete-orphan" inverse="true" fetch="join">
```
ctapobep: I think in most cases we don't need the contacts, do we?

oatkachenko: Yes, you're right. It was needed for plugins, when created copy of JCUser. I did research and made conclusion that we don't need contacts in this scope, so I remove copying of contacts and added javadocs with explanations.

jcommune-service/src/main/java/org/jtalks/jcommune/service/transactional/TransactionalUserService.java
```diff
@@ -82,6 +81,7 @@
     private final Base64Wrapper base64Wrapper;
     //Important, use for every password creation.
     private final EncryptionService encryptionService;
+    private SecurityContextFacade securityContextFacade;
```
ctapobep: `final`?

oatkachenko: oh it's my fault

PR #264, opened 2016-12-24
Fixed: smileys are displayed as rombuses in title of question and in comments to question

mihnayan:
JC-2350 Fixed: smileys (UTF-8) are displayed as rombuses in title
of the question

JC-2351 Fixed: smileys (UTF-8) are displayed as rombuses
in comments to the question

Fixed not correct escaping of Emoji in other elements of user
interface of QuestionsAndAnswers plugin.

To solve tasks used Unbescape Java library for correct escaping
strings that contain Emoji:  https://github.com/unbescape/unbescape

Next step: writing tests to check the correctness of text and title
in questions and answers.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -443,6 +443,12 @@
         <artifactId>spring-data-commons-core</artifactId>
         <version>1.3.0.RELEASE</version>
       </dependency>
+      <!-- Library for html5 escaping. Needed to correct escaping strings that contain Emoji -->
+      <dependency>
+        <groupId>org.unbescape</groupId>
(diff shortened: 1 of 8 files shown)


pom.xml
```diff
@@ -443,6 +443,12 @@
         <artifactId>spring-data-commons-core</artifactId>
         <version>1.3.0.RELEASE</version>
       </dependency>
+      <!-- Library for html5 escaping. Needed to correct escaping strings that contain Emoji -->
```
ctapobep: This is needed only in plugins as for now

PR #267, opened 2016-12-27
JC-2380 SpeedUpForum

oatkachenko:
Notes
When we need to get current user - performed inefficient query to the database as result we have performance issue.

TODO// describe all changes i've made.

Functionality Added or Changed
-SecurityContext keep UserInfo instead of JCUser
-SecurityService returns current user from SecurityContext as a result - less database queries

Bugs Fixed
- user not removed from LocationService#registerUserMap when session destroyed (memory leak).

### DIFF
--- a/jcommune-model/src/main/java/org/jtalks/jcommune/model/dao/UserDao.java
+++ b/jcommune-model/src/main/java/org/jtalks/jcommune/model/dao/UserDao.java
@@ -14,6 +14,7 @@
  */
 package org.jtalks.jcommune.model.dao;
 
+import org.hibernate.ObjectNotFoundException;
 import org.jtalks.common.model.entity.User;
 import org.jtalks.jcommune.model.entity.JCUser;
 
@@ -106,4 +107,12 @@ public interface UserDao extends org.jtalks.common.model.dao.UserDao<JCUser> {
      * @return list of found users
      */
     List<JCUser> findByUsernameOrEmail(String pattern, int count);
+
+    /**
+     * May return a proxy (without hitting DB). If no row found the object is returned but it
+     * throws {@link ObjectNotFoundException} when properties are first accessed.
+     * @param id stored user identifier.
+     * @return proxy of JCUser object.
+     */
+    JCUser loadById(Long id);
 }
--- a/jcommune-model/src/main/java/org/jtalks/jcommune/model/entity/UserInfo.java
+++ b/jcommune-model/src/main/java/org/jtalks/jcommune/model/entity/UserInfo.java
@@ -1,0 +1,123 @@
+/**
+ * Copyright (C) 2011  JTalks.org Team
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ */
+
+package org.jtalks.jcommune.model.entity;
+
+import org.jtalks.common.model.entity.User;
+import org.springframework.security.core.Authentication;
+import org.springframework.security.core.GrantedAuthority;
+import org.springframework.security.core.userdetails.UserDetails;
+
+import java.util.Collection;
+import java.util.Collections;
+
+/**
+ * Provides core user information which is later encapsulated. This class is immutable which is very important
+ * as it's going to be accessed from many threads. All the fields here are loaded only once into HTTP Session
+ * and don't ever change within that session.
+ *
+ * into {@link Authentication} objects.
+ *
+ * @author Oleg Tkachenko
+ */
+public class UserInfo implements UserDetails {
+    private final long id;
+    private final String uuid;
+    private final String username;
+    private final String password;
+    private final boolean enabled;
+
+    public UserInfo(User user) {
+        this.id = user.getId();
+        this.uuid = user.getUuid();
+        this.username = user.getUsername();
+        this.password = user.getPassword();
+        this.enabled = user.isEnabled();
+    }
+
+    public long getId() {
+        return id;
+    }
+
+    public String getUuid() {
+        return uuid;
+    }
+
+    @Override
+    public String getUsername() {
+        return username;
+    }
+
+    @Override
+    public String getPassword() {
+        return password;
+    }
+
+    /**
+     * Returns user's activation status in forum. After registration user has status "disabled" and receives an e-mail
+     * with activation link. When user confirm his e-mail address, status changed to "enabled". Only users with
+     * activation status "enabled" can login.
+     *
+     * @return return true if user is activated in forum and can login, otherwise false.
+     */
+    @Override
+    public boolean isEnabled() {
+        return enabled;
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (!(o instanceof UserInfo)) return false;
+        UserInfo userInfo = (UserInfo) o;
+        return getUuid().equals(userInfo.getUuid());
+    }
+
+    @Override
+    public int hashCode() {
+        return getUuid().hashCode();
+    }
+
+    /*
+     * Next methods from UserDetails interface, indicating that user can or can't authenticate.
+     * We don't need this functionality.
+     */
+
+    /**
+     * Used for role based access in spring security.
+     * Jcommune uses Domain Object Security (ACLs) so we don't have any roles.
+     *
+     * @return {@link Collections.EmptyList}
+     */
+    @Override
+    public Collection<GrantedAuthority> getAuthorities() {
+        return Collections.emptyList();
+    }
+
+    @Override
+    public boolean isAccountNonExpired() {
+        return true;
+    }
+
+    @Override
+    public boolean isAccountNonLocked() {
+        return true;
+    }
+
+    @Override
+    public boolean isCredentialsNonExpired() {
+        return true;
+    }
+}
--- a/jcommune-service/src/main/java/org/jtalks/jcommune/service/nontransactional/LocationService.java
+++ b/jcommune-service/src/main/java/org/jtalks/jcommune/service/nontransactional/LocationService.java
@@ -16,9 +16,9 @@
 package org.jtalks.jcommune.service.nontransactional;
 
 import org.jtalks.common.model.entity.Entity;
-import org.jtalks.jcommune.model.entity.JCUser;
+import org.jtalks.jcommune.model.entity.UserInfo;
 import org.jtalks.jcommune.plugin.api.service.PluginLocationService;
-import org.jtalks.jcommune.service.UserService;
+import org.jtalks.jcommune.service.security.SecurityService;
 import org.springframework.security.core.session.SessionRegistry;
 import org.springframework.stereotype.Component;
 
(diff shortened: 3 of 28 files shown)


jcommune-model/src/main/java/org/jtalks/jcommune/model/dao/UserDao.java
```diff
@@ -106,4 +106,11 @@
      * @return list of found users
      */
     List<JCUser> findByUsernameOrEmail(String pattern, int count);
+
+    /**
+     * Return a “proxy” of JCUser without hitting the database. If no row found, it will throws an ObjectNotFoundException.
```
ctapobep: _May_ return a proxy (without hitting DB). If no row found the object is returned but it throws `ObjectNotFoundException` when properties are first accessed.

jcommune-model/src/main/java/org/jtalks/jcommune/model/entity/UserInfo.java
```diff
@@ -0,0 +1,139 @@
...
+import org.springframework.security.core.Authentication;
+import org.springframework.security.core.GrantedAuthority;
+import org.springframework.security.core.userdetails.UserDetails;
+
+import java.util.Collection;
+import java.util.Collections;
+
+/**
+ * Provides core user information which is later encapsulated
+ * into {@link Authentication} objects.
```
ctapobep: `This class is immutable which is very important as it's going to be accessed from many threads. All the fields here are loaded only once into HTTP Session and don't ever change within that session.`

@oatkachenko have checked the functionality of changing the password?

jcommune-model/src/main/java/org/jtalks/jcommune/model/entity/UserInfo.java
```diff
@@ -0,0 +1,139 @@
...
+    }
+
+    @Override
+    public int hashCode() {
+        return getUuid().hashCode();
+    }
+
+    /*
+     * Next methods from UserDetails interface, indicating that user can or can't authenticate.
+     * We don't need this functionality yet so users always enabled.
```
ctapobep: > so users always enabled

So shouldn't we put `isEnabled()` here and make it always return `true`?

jcommune-service/src/main/java/org/jtalks/jcommune/service/nontransactional/LocationService.java
```diff
@@ -82,6 +82,19 @@ public LocationService(UserService userService, SessionRegistry sessionRegistry)
...
+    }
+
+    /**
+     * Clears forum location for the given user. After the call, user
+     * passed as param will be excluded from all the topic/branch viewer's list
+     *
+     * @param user to clean location.
+     */
+    public void clearUserLocation(Object user) {
+        if (user instanceof UserInfo) {
```
ctapobep: Do we need to check for `instanceof`? If it's not in the map - it's not in the map regardless of whether it's a `UserInfo` or not.

jcommune-service/src/main/java/org/jtalks/jcommune/service/security/acl/AclGroupPermissionEvaluator.java
```diff
@@ -248,13 +249,13 @@ private boolean isGrantedForSid(Sid sid, AccessControlEntry ace,
      */
     private boolean isGrantedPersonalPermission(Authentication authentication, Permission permission,
                                                 boolean isCheckAllowedGrant) {
-            JCUser storedUser = (JCUser) authentication.getPrincipal();
-            // retriev user with replicated groups from EhCache
-            JCUser actualUser = userDao.get(storedUser.getId());
+            UserInfo principal = (UserInfo) authentication.getPrincipal();
+            // retrieve user with replicated groups from EhCache
```
ctapobep: We don't have replicated cache anymore..

jcommune-service/src/main/java/org/jtalks/jcommune/service/security/acl/AclGroupPermissionEvaluator.java
```diff
@@ -248,13 +249,13 @@ private boolean isGrantedForSid(Sid sid, AccessControlEntry ace,
...
+            JCUser actualUser = userDao.get(principal.getId());
```
ctapobep: You need to use service here, don't you?

jcommune-service/src/test/java/org/jtalks/jcommune/service/security/AclGroupPermissionEvaluatorTest.java
```diff
@@ -81,21 +82,19 @@ public void init() throws NotFoundException {
                 sidFactory, mutableAclService, userDao, pluginManager);
         objectIdentity = new ObjectIdentityImpl(targetType, targetId);
         Mockito.when(aclUtil.createIdentity(targetId, targetType)).thenReturn(objectIdentity);
-        user = Mockito.mock(JCUser.class);
-        when(user.getId()).thenReturn(1L);
-        userSid = new UserSid(user);
+        user = new JCUser("username", "email@jtalks.org", "password");
+        user.setId(1L);
```
ctapobep: `Long()`?

jcommune-service/src/test/java/org/jtalks/jcommune/service/security/AclGroupPermissionEvaluatorTest.java
```diff
@@ -130,8 +129,8 @@ public void testHasPermissionForPermissionOnGroupSuccessTest() throws Exception
     public void testHasPermissionForPermissionOnGroupNonExistentUserTest() throws Exception {
         setEnvForPermissionOnGroupTests(true);
         Long nonExistentId = -1L;
-        when(user.getId()).thenReturn(nonExistentId);
-        when(user.getGroupsIDs()).thenReturn(new ArrayList<Long>());
+        user.setId(nonExistentId);
```
ctapobep: `Long()`

jcommune-view/jcommune-web-controller/src/main/java/org/jtalks/jcommune/web/listeners/SessionSetupListener.java
```diff
@@ -39,10 +47,19 @@
...
+        ServletContext servletContext;
+        ApplicationContext ctx;
+        JCommuneProperty property = null;
+        try {
+            servletContext = se.getSession().getServletContext();
+            ctx = WebApplicationContextUtils.getRequiredWebApplicationContext(servletContext);
+            property = (JCommuneProperty) ctx.getBean("sessionTimeoutProperty");
+            locationService = (LocationService) ctx.getBean("locationService");
+        } catch (Exception ex) {
+            logger.warn("Bean instantiation error: " + ex.getMessage());
```
ctapobep: Shouldn't we re-throw the error to make the app crash during startup?

jcommune-view/jcommune-web-view/src/test/java/org/jtalks/jcommune/test/utils/Users.groovy
```diff
@@ -20,6 +20,7 @@ import org.jtalks.common.service.security.SecurityContextFacade
 import org.jtalks.jcommune.model.dao.GroupDao
```
ctapobep: Do you want to write a component test that checks that users viewing or location service work correctly with sessions?

jcommune-model/src/main/java/org/jtalks/jcommune/model/entity/UserInfo.java
oatkachenko: No, in this case user can login even without activation on forum.

jcommune-service/src/main/java/org/jtalks/jcommune/service/nontransactional/LocationService.java
oatkachenko: if key is null - NullPointerException will be thrown, or am i wrong? I could check for null but I think it's more pretty.

jcommune-model/src/main/java/org/jtalks/jcommune/model/entity/UserInfo.java
oatkachenko: yes, changing password works as expected.

jcommune-service/src/main/java/org/jtalks/jcommune/service/security/SecurityService.java
```diff
@@ -134,6 +138,16 @@ public UserDetails loadUserByUsername(String username) {
         if (user == null) {
```
ctapobep: DAO is still used instead of Service here :)

jcommune-view/jcommune-web-controller/src/main/java/org/jtalks/jcommune/web/listeners/SessionSetupListener.java
```diff
@@ -39,10 +47,16 @@
...
-        ServletContext servletContext = se.getSession().getServletContext();
-        ApplicationContext ctx = WebApplicationContextUtils.getRequiredWebApplicationContext(servletContext);
-        JCommuneProperty property = (JCommuneProperty) ctx.getBean("sessionTimeoutProperty");
-        int timeoutInSeconds = (int) TimeUnit.SECONDS.convert(property.intValue(), TimeUnit.MINUTES);
+        JCommuneProperty property;
+        ServletContext context = se.getSession().getServletContext();
+        try {
+            property = getBeanFormApplicationContext(context, JCommuneProperty.class, "sessionTimeoutProperty");
+        } catch (Exception ex) {
+            logger.warn("Bean instantiation error: " + ex.getMessage());
```
ctapobep: `logger.warn("Bean instantiation error.", ex.getMessage());`?

jcommune-view/jcommune-web-controller/src/main/java/org/jtalks/jcommune/web/listeners/SessionSetupListener.java
```diff
@@ -58,6 +72,43 @@ public void sessionCreated(HttpSessionEvent se) {
      */
     @Override
     public void sessionDestroyed(HttpSessionEvent se) {
-        //noop
+        LocationService locationService;
+        ServletContext context = se.getSession().getServletContext();
+        Object principal = getPrincipalFromSession(se.getSession());
+        if (principal != null) {
+            try {
+                locationService = getBeanFormApplicationContext(context, LocationService.class, "locationService");
```
ctapobep: Is it possible to retrieve that bean only once? I'm not sure how performant this operation is in Spring..

jcommune-view/jcommune-web-view/src/main/webapp/WEB-INF/tags/users.tag
```diff
@@ -25,18 +25,6 @@
   <c:when test="${!(empty users)}">
     <c:forEach var="user" items="${users}" varStatus="i">
       <c:set var='labelClass' value=''/>
-      <c:if test="${branch != null}">
```
ctapobep: Don't forget to write a comment in the task that this functionality was removed.

jcommune-service/src/main/java/org/jtalks/jcommune/service/nontransactional/LocationService.java
```diff
@@ -35,16 +35,16 @@
  */
 @Component
```
ctapobep: Hm.. Why is this marked as `@Component` o_O

jcommune-service/src/main/java/org/jtalks/jcommune/service/nontransactional/LocationService.java
```diff
@@ -35,16 +35,16 @@
...
 public class LocationService implements PluginLocationService {
-    private UserService userService;
-    private SessionRegistry sessionRegistry;
-    private Map<JCUser, String> registerUserMap = new ConcurrentHashMap<>();
+    private final SecurityService securityService;
+    private final SessionRegistry sessionRegistry;
+    private Map<UserInfo, String> registerUserMap = new ConcurrentHashMap<>();
```
ctapobep: Can also be `final`

jcommune-service/src/main/java/org/jtalks/jcommune/service/nontransactional/LocationService.java
```diff
@@ -35,16 +35,16 @@
...
 
     /**
-     * @param userService     to figure out the current user
+     * @param securityService for retrieving basic info about current user.
```
ctapobep: Dots are not placed in the `@param` unless there are more than 1 sentence

jcommune-view/jcommune-web-controller/src/main/java/org/jtalks/jcommune/web/listeners/SessionSetupListener.java
```diff
@@ -28,36 +36,101 @@
...
+                        /* Zero property value should mean infinitive session.
+                        To disable session expiration we should pass negative value here.
+                        Setting zero as is results in weird Tomcat behavior.*/
+                        TIME_OUT_SECONDS = seconds > 0 ? seconds : -1;
+                    }
+                }
+            }
+            se.getSession().setMaxInactiveInterval(TIME_OUT_SECONDS);
+        } catch (Exception ex) {
+            logger.warn("Bean instantiation error: " + ex.getMessage());
```
ctapobep: logger.warn("...", ex).

jcommune-view/jcommune-web-view/src/test/java/org/jtalks/jcommune/test/LocationServiceTest.groovy
```diff
@@ -0,0 +1,95 @@
...
+            def branch = branches.created()
+        and: "User registered and has permissions to view topics in current branch"
+            users.created(user).withPermissionOn(branch, BranchPermission.VIEW_TOPICS)
+        and: "User logged in"
+            def session = users.signIn(user)
+        when: 'user open the branch page'
+            branches.open(branch, session)
+        then: 'user is present in viewers list'
+            def viewing = locationService.getUsersViewing(branch)
+            assertTrue(viewing.find().username == user.username)
```
ctapobep: assertEquals()?

oatkachenko:
fixed

PR #269, opened 2017-01-07
JC-2381 Controller exception to HTTP errors mapping without 302 redirection

usharik:


ctapobep:
Could you please squash commits into 1 so that I merge them?

ctapobep:
Merged

PR #275, opened 2017-01-19
JC-2367 Edit user groups

oatkachenko:
-fixed XSS vulnerability through group description field


ctapobep:
Could you please change the 1st line of the commit message to be more expressive? E.g. `JC-2367 Fix XSS vulnerability for group listing`

PR #278, opened 2017-01-30
JC-2389 Email black-list

oatkachenko:
- Added ability to block new user registration, if email address in the black list


jcommune-model/src/main/java/org/jtalks/jcommune/model/entity/JCommuneProperty.java
```diff
@@ -103,8 +103,9 @@
     /** title prefix - should be displayed at the beginning of the title of the every page */
     ALL_PAGES_TITLE_PREFIX,
     
-    COPYRIGHT
-    ;
+    COPYRIGHT,
+    /** List of email domains restricted to register with*/
+    EMAIL_BLACK_LIST;
```
masyan: chnage all variable   email -> emailDomains

oatkachenko: fixed

PR #281, opened 2017-02-07
JC-2351 Fixed: Smileys are displayed as rombuses in comments

mihnayan:
Using custom processor for escaping of HTML because standard Velocity
escaping utility not correct displays emoji.

Changed version of kefirbb library to 1.5


ctapobep:
Merged!

PR #288, opened 2017-02-19
JC-963 An incorrect message appears when a user performs the 'Remove …

keik13:
…avatar' action for an empty avatar.

Added disable button "Remove avatar", if the avatar default.
Checks 2 cases:
 1. when a default avatar has never changed.
 2. when a user downloads your avatar, but then deleted it and it changed
   to the default avatar.
Avatars in these cases differ by their size, etc.


ctapobep:
Hi, @keik13 I thought more on this and I think it's not that important for the end user but in the same time the performance is going to be a bit worse. So my inclination would be not to merge this. Sorry for the misleading JIRA state..

PR #292, opened 2017-03-22
clean up thread-locals

oatkachenko:
Tomcat can't remove JCUser objects from thread-local variables after application stop, so let's remove them after each Http request.


jcommune-service/src/main/java/org/jtalks/jcommune/service/security/SecurityService.java
```diff
@@ -43,7 +43,7 @@
     private final AclManager aclManager;
     private final AclBuilders aclBuilders = new AclBuilders();
     private final SecurityContextFacade securityContextFacade;
-    private ThreadLocal<JCUser> cachedUser = new ThreadLocal<>();
+    private final static ThreadLocal<JCUser> cachedUser = new ThreadLocal<>();
```
ctapobep: Shouldn't it be CACHED_USER in this case?

oatkachenko:
fixed

PR #300, opened 2017-04-12
JC-2390 minor fixes

oatkachenko:
- checkbox and icon now are on the same line
- removed tooltips for Enabled checkbox activation/deactivation
- fixed null-pointer exception in SpamProtectionService when email is null
- spaces are completely trimmed in "regex" and "description" fields
- fixed typo in spam-rule management dialog
- added word wrapping for regex and description columns
- added regex pattern validation
- added test for regex pattern validation
- added pop-up with 'Access Denied' message on spam protection page
- updated Spring Security to 3.1.1 This was necessary to facilitate the
  process of managing access to API


jcommune-view/jcommune-web-controller/src/main/java/org/jtalks/jcommune/web/controller/api/SpamRulesApi.java
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    @ResponseBody @RequestMapping(method = RequestMethod.POST)
+    public JsonResponse add(@Valid @RequestBody SpamRuleDto ruleDto, BindingResult result) throws org.jtalks.common.service.exceptions.NotFoundException, InterruptedException {
+        checkForAdminPermissions();
+        trimAndValidate(ruleDto, result);
+        if (result.hasFieldErrors() || result.hasGlobalErrors()) {
+            return new JsonResponse(JsonResponseStatus.FAIL, result.getAllErrors());
+        }
+        SpamRule spamRule = SpamRule.toEntity(ruleDto);
```
ctapobep: Shouldn't this be `ruleDto.toEntity()`?

jcommune-view/jcommune-web-controller/src/main/java/org/jtalks/jcommune/web/filters/ForceGetIfAccessDeniedFilter.java
```diff
@@ -0,0 +1,61 @@
...
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.List;
+
+import static org.springframework.http.HttpStatus.FORBIDDEN;
+
+/**
+ * @author Oleg Tkachenko
```
ctapobep: Please add JavaDocs explaining what this does and why we need it

jcommune-view/jcommune-web-view/src/main/webapp/WEB-INF/web.xml
```diff
@@ -151,6 +151,16 @@
         <url-pattern>/*</url-pattern>
     </filter-mapping>
 
+    <!--If request method to login page isn't GET or POST, filter changes it to GET-->
```
ctapobep: Why?

oatkachenko: It is needed to avoid getting http status `405-Method not allowed`. According to security-context.xml only authenticated users can have access to `/api/spam-rules`, so when Anonymous user performs [PUT] request to `/api/spam-rules` it is redirected to `/login` but there is no mapping specified to serve [PUT] request and we recieve response with status 405. Last time we discussed about this case and you suggested to add another `<http>` tag for [PUT] request in security-context.xml but I didn't find solution. If my solution is unacceptable, can you give me a clue?

jcommune-view/jcommune-web-controller/src/main/java/org/jtalks/jcommune/web/controller/api/SpamRulesApi.java
oatkachenko: I agree it looks a bit weird.

jcommune-view/jcommune-web-view/src/main/webapp/WEB-INF/web.xml
ctapobep: By "why" I meant "please left a comment in source code to explain these lines" :) 

A better solution could be to map this or a different filter to `/api/*`

ctapobep:
Merged

PR #305, opened 2017-06-01
JC-2400 Service to get user out of appropriate group

usharik:
with name and email matching pattern

### DIFF
--- a/jcommune-model/src/main/resources/org/jtalks/jcommune/model/entity/JCUser.hbm.xml
+++ b/jcommune-model/src/main/resources/org/jtalks/jcommune/model/entity/JCUser.hbm.xml
@@ -108,4 +108,25 @@
                          LOWER(USERNAME) ]]>
   </query>
 
+  <query name="searchByEmailOrUsernameNotInGroupId">
+    <![CDATA[SELECT new org.jtalks.jcommune.model.dto.UserDto(u.username, u.email, u.id)
+             FROM User u
+             LEFT JOIN u.groups g
+            WHERE (g.id != :groupId OR g.id IS NULL) AND
+                  (LOWER(u.username) LIKE (:pattern) ESCAPE '|' OR
+                   LOWER(u.email) LIKE (:pattern) ESCAPE '|')
+            ORDER BY     CASE
+                          WHEN LOWER(u.username)=:exactMatch OR LOWER(u.email)=:exactMatch THEN 0
+                          WHEN LOWER(u.username) LIKE (:primaryPattern) ESCAPE '|' OR LOWER(u.email) LIKE (:primaryPattern) ESCAPE '|' THEN 1
+                          WHEN LOWER(u.username) LIKE (:thirdaryPattern) ESCAPE '|' OR LOWER(u.email) LIKE (:thirdaryPattern) ESCAPE '|' THEN 3
+                          WHEN LOWER(u.username) LIKE (:secondaryPattern) ESCAPE '|' OR LOWER(u.email) LIKE (:secondaryPattern) ESCAPE '|' THEN 2
+                         END,
+                         CASE
+                          WHEN LOWER(u.username) LIKE (:pattern) ESCAPE '|' AND LOWER(u.email) LIKE (:pattern) ESCAPE '|' THEN 0
+                          WHEN LOWER(u.username) LIKE (:pattern) ESCAPE '|' THEN 1
+                          WHEN LOWER(u.email) LIKE (:pattern) ESCAPE '|' THEN 2
+                         END,
+                         LOWER(u.username) ]]>
+  </query>
+
 </hibernate-mapping>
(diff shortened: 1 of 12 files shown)


jcommune-view/jcommune-web-controller/src/main/java/org/jtalks/jcommune/web/controller/AdministrationController.java
```diff
@@ -332,6 +322,16 @@ public ModelAndView getSpamProtectionPage(){
         return new ModelAndView("spamProtection").addObject("rules", ruleDtos);
     }
 
+    @RequestMapping(value = "/usersgroup/list", method = RequestMethod.GET)
```
ctapobep: In Spam Rules we use `/` to get the list of resources, I'd follow those conventions. And we're looking for users here, not for usergroups, right? So I suppose it should be: `/user`

jcommune-view/jcommune-web-controller/src/main/java/org/jtalks/jcommune/web/controller/AdministrationController.java
```diff
@@ -332,6 +322,16 @@ public ModelAndView getSpamProtectionPage(){
...
+    @ResponseBody
+    public JsonResponse findUserOutOfGroup(@RequestParam("notInGroupId") long notInGroupId,
+                                           @RequestParam("pattern") String searchRequest) {
```
ctapobep: Cannot allow empty pattern, otherwise the whole DB will be loaded.

jcommune-view/jcommune-web-view/src/test/java/org/jtalks/jcommune/test/AdministrationControllerTest.groovy
```diff
@@ -213,4 +221,37 @@ class AdministrationControllerTest extends Specification {
             groups.assertGroupUserPage(page2, groupUserList.subList(20, 40))
             groups.assertGroupUserPage(page3, groupUserList.subList(40, 45))
     }
+
+    def 'in searchlist should be users not in group with id and with username or email matching pattern'() {
```
ctapobep: This can be a much simpler test - it should just check that everything works together. The rest has to be tested on Dao level. 

Also it should check the validation if it exists on the controller level.

jcommune-view/jcommune-web-controller/src/main/java/org/jtalks/jcommune/web/controller/AdministrationController.java
usharik: We have a limit for users to show. Isn't it?

jcommune-view/jcommune-web-controller/src/main/java/org/jtalks/jcommune/web/controller/AdministrationController.java
usharik: Could you clarify which URL you expect here?

jcommune-model/src/main/resources/org/jtalks/jcommune/model/entity/JCUser.hbm.xml
```diff
@@ -108,4 +108,25 @@
...
+                          WHEN LOWER(u.username) LIKE (:primaryPattern) ESCAPE '|' OR LOWER(u.email) LIKE (:primaryPattern) ESCAPE '|' THEN 1
+                          WHEN LOWER(u.username) LIKE (:thirdaryPattern) ESCAPE '|' OR LOWER(u.email) LIKE (:thirdaryPattern) ESCAPE '|' THEN 3
+                          WHEN LOWER(u.username) LIKE (:secondaryPattern) ESCAPE '|' OR LOWER(u.email) LIKE (:secondaryPattern) ESCAPE '|' THEN 2
+                         END,
+                         CASE
+                          WHEN LOWER(u.username) LIKE (:pattern) ESCAPE '|' AND LOWER(u.email) LIKE (:pattern) ESCAPE '|' THEN 0
+                          WHEN LOWER(u.username) LIKE (:pattern) ESCAPE '|' THEN 1
+                          WHEN LOWER(u.email) LIKE (:pattern) ESCAPE '|' THEN 2
+                         END,
+                         LOWER(u.username) ]]>
```
usharik: No. Could you help me to get backup of production DB with sufficient count of users?
And another point that we already have another query very close to this one which normally works on production.

ctapobep: I see.. Maybe it's not a big deal. @oatkachenko has a backup.

jcommune-view/jcommune-web-controller/src/main/java/org/jtalks/jcommune/web/controller/AdministrationController.java
ctapobep: As I said it should be `/user?pattern=blah&notInGroup=1`

usharik: I think notInGroup=0 should be supported too. Isn't it?

ctapobep: If this should be supported, I'd rather make the parameter optional

usharik: Done. As we already have method mapped to `/user` endpoint I used mapping by parameters.

PR #309, opened 2017-09-26
JC-2400 Search users that should be added to usergroup

evgeniycheban:
- FAB tool tip is absent. Fixed
- Change user adding to group pop up UI and functionality. Fixed
- Open the autocomplete list after entering any character. Fixed
- Check field length error message is shown for 51 symbols entered. Fixed


oatkachenko:
Looks good, but I have one question. If user input only one character then error message displayed. Is it correct behaviour? I found only one mention about this case:
**Expected results:**

- The autocomplete list should not appear
- Sorting users in the autocomplete list should not start

Maybe, we should send request to server only after second character?

evgeniycheban:
@oatkachenko I think you are right. I will update it.

evgeniycheban:
Updated.

oatkachenko:
thx, @ctapobep I have no questions.

PR #310, opened 2017-12-10
JC-2409 Error 500 when saving PM with invalid "To" field

evgeniycheban:
- Fixed hibernate lazy initialization exception by moving
userService.getCurrentUser() to service layer.


ctapobep:
Can we write a test that would've failed before this change?

evgeniycheban:
@ctapobep yes, I'll write a test in this week.

